PR #4, opened 2019-08-07
region-k-means

sjsrey:
New algo that gives us k-connected components following k-means logic.

We may want to move some of the helper functions out of this module, but this gets us moving.


renanxcortes:
Looks goo to me. One question, are we gonna use some of the stuff present in https://github.com/pysal/region/blob/master/region/csgraph_utils.py?

Issue #11, opened 2019-08-22
update python versions, etc. in setup.py


jGaboardi:
* ~~Python 3.5, 3.6~~ --> Python 3.6, 3.7

* add [`requirements_pip.txt`](https://github.com/pysal/spopt/blob/master/requirements_pip.txt) to `setup_package()`

* use `package` variable for name

* add GitHub `url`

jGaboardi:
Addressed in #12

PR #12, opened 2019-08-23
Setup update

jGaboardi:
Addressing #11

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,8 @@ def setup_package():
           install_requires=install_reqs,
           extras_require=extras_reqs,
           zip_safe=False,
-          cmdclass = {'build.py':build_py})
+          cmdclass = {'build.py':build_py},
+          python_requires='>3.5')
 
 if __name__ == '__main__':
     setup_package()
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -21,17 +23,18 @@ def setup_package():
...
 
-    setup(name='spopt', #name of package
+    setup(name=package,
           version=__version__,
           description='', #short <80chr description
-          url='', #github repo
+          url='https://github.com/pysal/'+package, #github repo
           maintainer='',
           maintainer_email='',
           test_suite = 'nose.collector',
```
knaaptime: i dont know that it matters but we're using pytest rather than nose

knaaptime: or, we'd intended to anyway. Not sure an official decision has been made since we dont have any tests yet

jGaboardi: it probably does matter... i'll make the switch.

jGaboardi: @knaaptime would this simply be `test_suite = 'pytest'`?

jGaboardi:
OK, well, maybe we'll leave `test_suite = 'nose.collector'` as it is for now and change later once we decide on testing.

PR #22, opened 2020-04-19
streamlining GHA testing

jGaboardi:
Address [this comment](https://github.com/pysal/spopt/pull/20#pullrequestreview-396055194).

Also, we are generating a `cov` report, but not actually checking/comparing/reporting anything from it. If we would like to set up `codecov`, I can do that easily, I will just need admin access to `spopt`.


jGaboardi:
Going ahead to merge this and focus on the coverage stuff in #24

PR #53, opened 2020-05-12
adding customized codecov.yml

jGaboardi:
This PR adds a customized `codecov.yml` that:
* allows for a 2% change threshold without failing
* decreases precision to 1 decimal place
* assigns 95% coverage as "complete"
* waits until all 9 builds are complete before commenting here


jGaboardi:
Going to merge this today unless there are any objections.

Issue #61, opened 2020-08-14
solvers and solver APIs


jGaboardi:
#18 #49 

Following the [2020-08-14 dev meeting](https://hackmd.io/lEQiaEDgSvCjtqUutk8hew), it was decided that all solvers APIs will be conditional dependencies and will be available through a [`solver.py` module](https://hackmd.io/lEQiaEDgSvCjtqUutk8hew#Solver-binder). @sjsrey, can you provide a description and example on how that can be implemented?

Further, the distinction was clarified between a solver and a solver API, with examples of the former being [CBC](https://github.com/coin-or/Cbc) and [Gurobi](https://www.gurobi.com), and examples of the latter being [PuLP](https://github.com/coin-or/pulp) and [Python-MIP](https://github.com/coin-or/python-mip).

jGaboardi:
closing for now -- solver agnostic is the way to go -- we just have to make sure the include all necessary solvers/wrappers in the `requirements`, etc.

PR #62, opened 2020-08-14
removing python-mip language from REAMDE

jGaboardi:
See #61, etc.


jGaboardi:
I'll go ahead and merge this today unless there are any objections.

PR #65, opened 2020-09-01
skater

xf37:
add function skater and test_skater


spopt/region/skater.py
```diff
@@ -0,0 +1,342 @@
...
+           + df.filter(like='89').columns.tolist()].values
+    data_c = (data - data.mean(axis=0)) / data.std(axis=0)
+    W = weights.Queen.from_dataframe(df)
+    result = Spanning_Forest().fit(10, W, data_c, quorum=100)
+    
+    will_fail = Spanning_Forest().fit(10,W,data_c, quorum=500)
+
+
+
+class SKATER(BaseSpOptHeuristicSolver):
```
jGaboardi, line 285: Move this to above the `if __name__ == "__main__":` line.

xf37:
I got 2 failure checks about "cannot import name 'SKATER' from 'spopt'" when text_skater is running. Is it because the skater function has not been merged? Should I commit these two separately? Thanks in advance! @sjsrey

jGaboardi:
add `sklearn` to [`requirements.txt`](https://github.com/pysal/spopt/blob/master/requirements.txt)

jGaboardi:
These were the requirements for the original [`region`](https://github.com/pysal/region/blob/32861c36300e99e96ba15cfcc888d69a12046d2a/setup.py#L43) PySAL submodule. Also, it looks like `geopandas` is being used in the tests for the PR, but it is not in out current requirements. That should be added, as well.

requirements.txt
```diff
@@ -3,4 +3,7 @@ numpy>=1.3
 pandas>=1
 networkx
 libpysal
-ortools
\ No newline at end of file
+ortools
+sklearn
```
jGaboardi, line 7: Ah! This should be `scikit-learn`! (See[ here](https://scikit-learn.org/stable/install.html#installation-instructions)). I am always confused about `scikit-learn`'s install name and import alias...

xf37: Is it possible due to the [version of python](https://stackoverflow.com/questions/36404042/importerror-no-module-named-sklearn-python)?

jGaboardi: It may be, but I am still not sure why the environment seems to not even attempt to download `scikit-learn`. See environment list [here](https://github.com/pysal/spopt/pull/65/checks?check_run_id=1059112512#step:3:832). I went through the `Run goanpeca/setup-miniconda@v1` process and did not even see any mention of `scikit-learn`. However, `geopandas` (which you just added) is being installed. This seems to be strange, but it will most likely be something small we overlooked... I have to work on some other stuff now, but maybe @knaaptime can give some advice?

jGaboardi: [`conda-forge/scikit-learn`](https://github.com/conda-forge/scikit-learn-feedstock#current-build-status) is passing on all platforms, so it does not appear to be version/Python issue.

jGaboardi: > This seems to be strange, but it will most likely be something small we overlooked... 

This ended up being true. It was an oversight of mine... see fe8dbcd

spopt/__init__.py
```diff
@@ -3,4 +3,4 @@
 
 
 # import modules/functions
-from .region import MaxPHeuristic, RegionKMeansHeuristic, w_to_g
+from .region import MaxPHeuristic, RegionKMeansHeuristic, SKATER, w_to_g
```
jGaboardi, line 6: Should we also expose `Spanning_Forest` here and in `region/__init__.py`?

spopt/tests/test_skater.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
+import libpysal
+import geopandas as gpd
+from .. import SKATER
+
+def test_skater():
```
jGaboardi, line 6: IMHO we should be using the `unittest`-style testing (as in [`esda`](https://github.com/pysal/esda/blob/e66098341ca8c8b31ceff2025edeefbf10df023f/esda/tests/test_moran.py#L12)), but I will defer to @sjsrey's judgment.

jGaboardi:
There are also quite a few new lines of code in [`skater.py`](https://github.com/pysal/spopt/pull/65/checks?check_run_id=1061447495) that are not covered in [`test_skater.py`](https://github.com/pysal/spopt/pull/65/files#diff-a79f2c1c27338083576631e112a4d04a). We ideally should have more than one test, and have tests that are designed to fail in order to cover as many cases as possible (see example in [`spaghetti`](https://github.com/pysal/spaghetti/blob/be8fc7182da051ce20ea017c042cfcf134599bb6/spaghetti/tests/network_unittest_classes.py#L152)).

spopt/__init__.py
knaaptime: can we commit to pep8 from the get-go in this package? I'm having a tough time trying to convert some of the old packages (so `SpanningForest` if you want it) ?

knaaptime: also probably `Skater`...

jGaboardi: > can we commit to pep8 from the get-go in this package? I'm having a tough time trying to convert some of the old packages (so SpanningForest if you want it) ?

So do you mean:

```python
from region import MaxPHeuristic, RegionKMeansHeuristic, SKATER, w_to_g
```

instead of

```python
from .region import MaxPHeuristic, RegionKMeansHeuristic, SKATER, w_to_g
```
?

spopt/region/skater.py
```diff
@@ -0,0 +1,342 @@
...
+from scipy.sparse import csgraph as cg
+from scipy.optimize import OptimizeWarning
+from collections import namedtuple
+from warnings import warn
+import time
+import numpy as np
+import copy
+deletion = namedtuple('deletion', ('in_node', 'out_node', 'score'))
+
+class Spanning_Forest(object):
```
knaaptime, line 14: can we make this `SpanningForest` instead?

spopt/__init__.py
knaaptime: no i just mean all classes should be camel cased

jGaboardi: > no i just mean all classes should be camel cased

I really over-thought that one...

spopt/tests/test_skater.py
xf37: Thanks for pointing this out. I just followed the format of [test_maxp.py](https://github.com/pysal/spopt/blob/master/spopt/tests/test_maxp.py) and [test_region_k_means.py](https://github.com/pysal/spopt/blob/master/spopt/tests/test_region_k_means.py). I have not written this before. Do you know whether there are the unittest -style testing for these two methods? How could they pass the checks?  @jGaboardi

jGaboardi: @xf37 I don't mean to suggest that the way you wrote the test is wrong. It is a perfectly valid test. When I say "`unittest`-style", I simply mean with the `unittest` package in something like the format below. By creating a `TestSkater(unittest.TestCase)` instance and adding the `Skater` model as an attribute in `set_up()` we can reuse the model in more tests. In the current test we are only testing the values of `model.labels_`, but there are many of things we could (should?) be testing.

```python
import unittest

class TestSkater(unittest.TestCase):

    def set_up():
        self.pth = libpysal.examples.get_path("mexicojoin.shp")
        self.mexico = gpd.read_file(self.pth)
        self.mexico["count"] = 1
        self.attrs_name = [f"PCGDP{year}" for year in range(1950, 2010, 10)]
        self.w = libpysal.weights.Queen.from_dataframe(self.mexico)
        np.random.seed(123456)
        self.model = Skater(gdf=mexico, w=w, attrs_name=attrs_name)
        self.model.solve()

    def test_labels_(self):
        observed_labels = self.model.labels_
        known_labels = np.array(
            [
                0,
                0,
                1,
                2,
                2,
                1,
                1,
                1,
                1,
                1,
                3,
                2,
                1,
                1,
                1,
                1,
                4,
                1,
                1,
                1,
                1,
                1,
                0,
                0,
                0,
                1,
                1,
                1,
                1,
                0,
                1,
                1,
            ]
     )
     numpy.testing.assert_equal(observed_labels, known_labels)
```

spopt/region/skater.py
```diff
@@ -0,0 +1,409 @@
...
+            MSF[in_node, out_node] = 0
+            MSF.eliminate_zeros()
+            return (MSF, *cg.connected_components(MSF, directed=False))
+        raise OptimizeWarning(
+            "Score of the ({},{}) cut is inf, the quorum is likely not met!"
+        )
+
+
+class Skater(BaseSpOptHeuristicSolver):
+    """...Needs a short description..."""
```
jGaboardi, line 340: Let's go ahead and fill in the description here and the docstrings in `__init__`.

spopt/region/skater.py
```diff
@@ -0,0 +1,409 @@
...
+        self.n_clusters = n_clusters
+        self.floor = floor
+        self.trace = trace
+        self.islands = islands
+
+    def solve(self):
+        """Solve the region k-means heuristic."""
+        data = self.gdf
+        X = data[self.attrs_name].values
+        model = SpanningForest()
```
jGaboardi, line 391: Shouldn't users be able to pass in the keyword arguments to `SpanningForest.__init__()` here (e.g. `dissimilarity`)?

spopt/region/skater.py
```diff
@@ -0,0 +1,409 @@
...
+        model.fit(self.n_clusters, self.w, data=X, quorum=self.floor, trace=self.trace)
```
jGaboardi, line 392: Users should also be able to pass in keywords for `SpanningForest.fit()`  here.

spopt/tests/test_skater.py
xf37: Gotcha, thank you for explaining.

xf37:
@jGaboardi Thank you for reviewing the PR! I have fixed all github CI test errors and would like to merge this PR. I may add more unit tests later.

spopt/region/skater.py
jGaboardi: We should be passing in the keywords more programmatically. Something like the following, for example:

```python

class Skater(BaseSpOptHeuristicSolver):
    def __init__(..., ...., ...., spanning_forest_kwds=dict()):
        
        self.spanning_forest_kwds = spanning_forest_kwds
        
    
    def solve(self):
        """Solve the region k-means heuristic."""
        
        model = SpanningForest(**self.spanning_forest_kwds)
         

```

Then `Skater` can be instantiated like so:

```python
kwargs = {"center": numpy.median}
Skater(gdf=mexico, w=w, attrs_name=attrs_name, spanning_forest_kwds=kwargs)
```

spopt/region/skater.py
```diff
@@ -0,0 +1,432 @@
...
+    def solve(self):
+        """Solve the region k-means heuristic."""
+        data = self.gdf
+        X = data[self.attrs_name].values
+        model = SpanningForest(self.dissimilarity, self.affinity, self.reduction, self.center)
+        model.fit(self.n_clusters, self.w, data=X, quorum=self.floor, trace=self.trace)
+        self.labels_ = model.current_labels_
+
+
+if __name__ == "__main__":
```
jGaboardi, line 419: I am not sure there is a reason for having this last section since the tests are happening in `tests/test_skater.py`. Is there a reason for keeping it that I am not seeing?

jGaboardi (changes requested):
I still have some concerns regarding the code before this gets merged (and we might want to double check if @sjsrey would like to give his review). I also still have concerns about the docstrings, but that can be beefed up and polished in a later PR.

spopt/region/skater.py
xf37: Thanks, I just removed it.

jGaboardi:
@xf37 Very nicely done! Let's wait until the end of the day if @sjsrey has time for a review then I'll merge this evening.

jGaboardi (approved):

PR #67, opened 2020-09-16
add spenc and test_spenc

xf37:


xf37:
Could you please help me check why the solution got by checks is different from what I got from my local machine? I used the same code and have specified random seed. @jGaboardi

ljwolf:
i think you need to use the `random_seed=` argument rather than setting `numpy.random.seed`, since [`SPENC` builds on a scikit-learn estimator with that option.](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.SpectralClustering.html)

jGaboardi:
> i think you need to use the `random_seed=` argument rather than setting `numpy.random.seed`, since [`SPENC` builds on a scikit-learn estimator with that option.](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.SpectralClustering.html)

I have the same thought.

spopt/tests/test_spenc.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+    def setUp(self):
+        self.pth = libpysal.examples.get_path("mexicojoin.shp")
+        self.mexico = gpd.read_file(self.pth)
+        self.mexico["count"] = 1
+        self.attrs_name = [f"PCGDP{year}" for year in range(1950, 2010, 10)]
+        self.w = libpysal.weights.Queen.from_dataframe(self.mexico)
+        np.random.seed(123456)
+        self.model = Spenc(gdf=self.mexico, w=self.w, attrs_name=self.attrs_name)
```
jGaboardi, line 18: Try adding the random state as a keyword to `Spenc()` that is then passed into `SPENC()`

xf37:
> i think you need to use the `random_seed=` argument rather than setting `numpy.random.seed`, since [`SPENC` builds on a scikit-learn estimator with that option.](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.SpectralClustering.html)

Thanks Levi!

spopt/region/spenc.py
```diff
@@ -9,7 +9,7 @@ class Spenc(BaseSpOptHeuristicSolver):
 	Spatially encouraged spectral clustering.
 	:cite:`wolf2018`
 	"""
-	def __init__(self, gdf, w, attrs_name, n_clusters=5, gamma=1):
+	def __init__(self, gdf, w, attrs_name, n_clusters=5, random_state, gamma=1):
```
jGaboardi, line 12: `random_state` should probably be passed in as a keyword.

```python
def __init__(self, gdf, w, attrs_name, n_clusters=5, gamma=1, random_state=None):
```

spopt/tests/test_spenc.py
```diff
@@ -22,38 +22,38 @@ def test_labels_(self):
...
-                1,
-                1,
-                3,
-                1,                
-                4,
-                1,
-                1,
-                0,
-                0,
+                4
```
jGaboardi, line 25: Do we know that this is the correct solution?

xf37: I do not know whether it is correct. just want to make sure the solution is the same with one random seed.

jGaboardi: OK, then we'll definitely need a review from @sjsrey and @ljwolf on this.

xf37:
@jGaboardi  Could you please check this error? It seems we need to update any usage of the set-env and add-path workflow commands, while I do not know where to start. 
`Add "/usr/share/miniconda3/condabin" to PATH
Error: Unable to process command '::add-path::/usr/share/miniconda3/condabin' successfully.
Error: The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
Set 'CONDA="/usr/share/miniconda3"'
Error: Unable to process command '::set-env name=CONDA::/usr/share/miniconda3' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
/usr/share/miniconda3/condabin/conda config --add pkgs_dirs /home/runner/conda_pkgs_dir
Error: Unable to process command '::set-env name=CONDA_PKGS_DIR::/home/runner/conda_pkgs_dir' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/`

jGaboardi:
@xf37 I am having trouble pushing to this branch for some reason. Try swapping out `- uses: conda-incubator/setup-miniconda@v2` for `- uses: goanpeca/setup-miniconda@v1` on line 20 in `spopt/.github/workflows/unittests.yml`. Last week the GHA syntax got updated and the previous version of `setup-miniconda` that we were using switched repos. See pysal/spaghetti#555, pysal/libpysal#359  for background.

jGaboardi:
@xf37 We have two problems here, (1) is an easy fix and the other (2) is more involved.

1. We are moving away from Python 3.6 with the initial phase of dropping support for Windows. This is due directly to [failures upstream with `fiona`](https://github.com/pysal/spopt/pull/67/checks?check_run_id=1456279597#step:8:360), which easily remedied by [excluding Python 3.6/Windows tests](https://github.com/pysal/libpysal/pull/363).
2. To my knowledge, we have still not verified the result from testing ([here](https://github.com/pysal/spopt/pull/67#pullrequestreview-501308746)) is correct. Verifying this is critical prior to merge. Maybe we can reach out to Serge, Ran, Levi, etc. to confirm this testing result?

xf37:
> @xf37 We have two problems here, (1) is an easy fix and the other (2) is more involved.
> 
> 1. We are moving away from Python 3.6 with the initial phase of dropping support for Windows. This is due directly to [failures upstream with `fiona`](https://github.com/pysal/spopt/pull/67/checks?check_run_id=1456279597#step:8:360), which easily remedied by [excluding Python 3.6/Windows tests](https://github.com/pysal/libpysal/pull/363).
> 2. To my knowledge, we have still not verified the result from testing ([here](https://github.com/pysal/spopt/pull/67#pullrequestreview-501308746)) is correct. Verifying this is critical prior to merge. Maybe we can reach out to Serge, Ran, Levi, etc. to confirm this testing result?

Thanks @jGaboardi! I would like to solve the first problem and leave the second problem as it is for now. I just talked about it with Serge and Eli. They are okay with this test file which has the same idea as it is in geosnap for now.

jGaboardi (approved):

Issue #68, opened 2020-10-18
rename master branch to main branch


jGaboardi:
#### 1. Renaming the organization remote `master` branch.
1. Create a renamed copy of the `master` branch.
  * ```
     $ git branch -m master main
      ```
2. Push the renamed copy to ***the organization*** remote.
  * ```
     $ git push -u upstream main
     ```
3. Update the [`default`](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/changing-the-default-branch) branch in `Settings/Branches`.
4.  Delete ***the organization*** remote `master`.
  * ```
     $ git push upstream --delete master
     ```

-------------
#### 2. Renaming your remote `master` branch.
1. Create a renamed copy of the `master` branch.
  * ```
     $ git branch -m master main
      ```
2. Push the renamed copy to ***your*** remote.
  * ```
     $ git push -u origin main
     ```
3. Update the [`default`](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/changing-the-default-branch) branch in `Settings/Branches`.
4.  Delete ***your*** remote `master`.
  * ```
     $ git push origin --delete master
     ```

jGaboardi:
@xf37 I will take care of this switch. Following the switch in `pysal/spopt` rename your repo branch from `master` to `main` (`xf37/spopt - master` --> `xf37/spopt - main`) and push upstream to the `main` branch ***not*** the `master` branch.

jGaboardi:
Addressing in #74

PR #74, opened 2020-11-26
Updating links from `master` branch to `main` branch

jGaboardi:
In order to provide a more inclusive environment and address problematic language we should update the `master` branch to `main`. 

See #68, pysal/spaghetti#543, pysal/libpysal#353


jGaboardi:
@xf37 This is a minor PR, but I'd like to get it merged as soon as possible. If you have any feedback on it, please let me know.

jGaboardi:
Still not sure what's causing the Ubuntu builds for fail currently...

jGaboardi:
@xf37 Since there is no actual code in this PR, I think we should merge. If I don't hear back from you by tomorrow, I'll ahead with that.

Issue #79, opened 2020-12-07
An issue about "module 'spopt' has no attribute 'version'"


xf37:
When I ran `sphinx-build -b html docsrc docs` to build docs website, I got
```
Running Sphinx v3.1.2
Configuration error:
There is a programmable error in your configuration file:
Traceback (most recent call last):
File "/Users/anaconda3/lib/python3.8/site-packages/sphinx/config.py", line 319, in eval_configfile
execfile(filename, namespace)
File "/Users/anaconda3/lib/python3.8/site-packages/sphinx/util/pycompat.py", line 88, in execfile_
exec(code, _globals)
File "/Users/xinfeng/spopt_rst/docsrc/conf.py", line 60, in <module>
version = spopt.version
AttributeError: module 'spopt' has no attribute 'version'
```
However, the first line of the init.py under the folder spopt is:
`__version__ = "0.1.0"`

I have tried to type `import spopt` and `spopt.__version__` in python, and get the right answer. NOT sure why conf.py cannot obtain the version info via `import spopt`.

jGaboardi:
> When I ran `sphinx-build -b html docsrc docs`

Try running `make html` from within the `./docsrc` directory then `make sync` if that succeeds.

xf37:
> > When I ran `sphinx-build -b html docsrc docs`
> 
> Try running `make html` from within the `./docsrc` directory then `make sync` if that succeeds.

I tried while met the same error.

jGaboardi:
OK, I will try to take a look at this soon.

jGaboardi:
@xf37 Which version of `spenc` are you using? When I run `make html` I am getting the following?

```
Running Sphinx v3.3.1
/Users/the-gaboardi/miniconda3/envs/py3_spopt/lib/python3.8/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.cluster.spectral module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.cluster. Anything that cannot be imported from sklearn.cluster is now part of the private API.
  warnings.warn(message, FutureWarning)

Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/Users/the-gaboardi/miniconda3/envs/py3_spopt/lib/python3.8/site-packages/sphinx/config.py", line 319, in eval_config_file
    execfile_(filename, namespace)
[...]

PR #83, opened 2020-12-09
update .rst files

xf37:


docsrc/conf.py
```diff
@@ -7,11 +7,12 @@
 import sys, os
 import sphinx_bootstrap_theme
 
-sys.path.insert(0, os.path.abspath("../../"))
```
jGaboardi, line 10: I'm not sure this is the correct fix. In [`spaghetti`](https://github.com/pysal/spaghetti/blob/d3cbda62ed6cdfd2b6a4fb9bdfcb084610e6031b/docsrc/conf.py#L15), [`giddy`](https://github.com/pysal/giddy/blob/458c06c9346eded33bfa8f3b4de0f40bc467a904/docsrc/conf.py#L23), etc. this path is correct. Maybe we can get @weikang9009 to look at this if she has a moment?

jGaboardi:
We generally don't add the build files from the `./docsrc/` directory, only the `./docs` directory. See [`giddy/docsrc/`](https://github.com/pysal/giddy/tree/master/docsrc) as an example.

docsrc/conf.py
xf37: Serge helped me find the issue and fixed the problem.

xf37:
> We generally don't add the build files from the `./docsrc/` directory, only the `./docs` directory. See [`giddy/docsrc/`](https://github.com/pysal/giddy/tree/master/docsrc) as an example.

Got it. It is just the first version of the draft. Many details will be updated later.

docsrc/conf.py
```diff
@@ -160,7 +161,7 @@
 # -- Options for HTMLHelp output ------------------------------------------
 
 # Output file base name for HTML help builder.
-htmlhelp_basename = project + "doc"
```
jGaboardi, line 163: Is there a reason for not using the `project` variable here and elsewhere in `conf.py` (line 187, 195, etc...)? `project` is defined above in line 51 and can be used throughout `conf.py`. This allows for robust programming.

docsrc/conf.py
```diff
@@ -251,40 +258,3 @@ def setup(app):
     "scipy": ("https://docs.scipy.org/doc/scipy/reference/", None),
 }
 
-
```
jGaboardi, line 254: Why is this section being deleted? This will be needed for converting notebooks into documentation. Granted, we don't have any notebooks yet, but we will in the future. Best to keep it.

docsrc/conf.py
jGaboardi: It seems that it fixed the problem locally on your machine, but there may be an issue with your directory paths because `sys.path.insert(0, os.path.abspath("../../"))` works for all other PySAL submodules and it works for me locally with `spopt` (aside from the [`calinski_harabasz_score` error](https://github.com/pysal/spopt/issues/79), which is actually an unrelated problem). Best we figure out what's going on with your local build rather than changing this in `main`.

xf37:
This PR is closed since I revised it to be another PR "update docs and docscr".

Issue #110, opened 2021-01-13
various tasks for package admin/cleanup


jGaboardi:
* rename `LICENSE` -> `LICENSE.txt` 61353f8
* add a `MANIFEST.in` file 267753c
* update supported Python versions (+3.9, -3.6)
  * `setup.py` (also `python_requires=">=3.7"`) 21f963e
  * `installation.rst` 75a77b1
* update binder link creation in `conf.py` and DRY in general 89631cd
* remove pin spaces in `.ci/*.yml` files c958bf4
* reduce min builds before reporting in [`codecov.yml`](https://github.com/pysal/spopt/blob/1609f377226a2d95a2cc24026ba355e9b0fdb682/codecov.yml#L3) 3f6264f
* adjust `environment.yml` f788d0b
  * bump Python version to 3.9, -3.8
  * replace `coveralls` with `codecov`
* `requirements*.txt` files
  * alphabetize 89a1234
  * create `requirements_tests.txt` with `pytest`, `pytest-cov`, & `codecov` a10dbfc
  * create `requirements_dev.txt` with `black` & `pre-commit` c2da3e4
* update `setup.py` 21f963e
  *  add keyword in [`setup()`](https://github.com/pysal/spopt/blob/1609f377226a2d95a2cc24026ba355e9b0fdb682/setup.py#L33): `py_modules=[package]`
  * add dev & testing reqs to [`_groups_files`](https://github.com/pysal/spopt/blob/1609f377226a2d95a2cc24026ba355e9b0fdb682/setup.py#L25)

jGaboardi:
All resolved in #111

PR #111, opened 2021-01-13
Polish up and minor docs/reqs fixes

jGaboardi:
Address all points in #110.


environment.yml
```diff
@@ -2,17 +2,20 @@ name: spopt
...
   - geopandas>=0.7.0
+  - libpysal
   - matplotlib
+  - networkx
+  - numpy>=1.3
+  - pandas>=1
+  - pulp
   - scikit-learn>=0.22
+  - scipy>=1.0
+  # testing, etc.
```
knaaptime, line 15: the environment.yml is what defines the dependencies for the conda package, so i'd prefer to leave the testing and dev requirements out of it so we dont force casual users to install them if they're using conda

knaaptime: and docs

jGaboardi: Fair enough. I'll remove.

jGaboardi: resolved in 5a44d17

environment.yml
```diff
@@ -2,17 +2,16 @@ name: spopt
...
   - pytest
```
knaaptime: these too?

jGaboardi:
The failure due to the coverage decrease can be ignored... for now. However, we are desperately in need of a more complete/robust testing suite. At the moment only ***half*** our code base is being tested, which paves the way for undetected bugs.

jGaboardi:
@knaaptime Sorry, I missed those two `pytest` (652d2c1)

knaaptime (approved):

.ci/39.yaml
```diff
@@ -2,24 +2,24 @@ name: test
...
   - pytest
   - pytest-cov
-  - pulp
-  - coverage
-  - coveralls
   - sphinx>=1.4.3
   - sphinxcontrib-bibtex==1.0.0
   - sphinx_bootstrap_theme
-  - numpydoc
-  - nbsphinx
```
jGaboardi, line 24: Don't we need to leave all the doc stuff in here since this environments used to build the docs in [`build_docs.yml`](https://github.com/pysal/spopt/blob/1609f377226a2d95a2cc24026ba355e9b0fdb682/.github/workflows/build_docs.yml#L16).

knaaptime: yeah, the yamls in `.ci/` need to be overly complete (`requirements_[all]`), the `environment.yml` should mirror `requirements.txt`

i think its just a formatting thing on the diff though... they're still in there, they just got moved up

jGaboardi: > i think its just a formatting thing on the diff though... they're still in there, they just got moved up

Ah, correct. That was my oversight.

PR #125, opened 2021-01-16
Narrative for notebook and handling a single attribute.

sjsrey:


jGaboardi (approved):

environment.yml
```diff
@@ -11,6 +11,8 @@ dependencies:
   - numpy>=1.3
   - pandas>=1
   - pulp
+  - scikit-learn>=0.22
+  - scipy>=1.0
   - coverage
   - coveralls
```
jGaboardi, line 17: We don't need `coveralls` since we are using `codecov`. However, we probably don't want `codecov` here either since it is not needed to run the binders/notebooks, as @knaaptime points out [here](https://github.com/pysal/spopt/pull/111#discussion_r556647874).

jGaboardi: `descartes` is also in `environment.yml` twice.

knaaptime (approved):

PR #145, opened 2021-02-08
revised notebook for azp

xf37:


jGaboardi (approved):

xf37:
@jGaboardi Could you please leave "merging commit" to me after approving the change next time? I recently checked the contributor board (https://github.com/pysal/spopt/graphs/contributors) and found I only had 11 commits and added 50 lines of code. I just realized that all of these credits will be account on the person who merges the commit. Thanks :)

Issue #112, opened 2021-01-13
Improve testing suite


jGaboardi:
> The failure due to the coverage decrease can be ignored... for now. However, we are desperately in need of a more complete/robust testing suite. At the moment only ***half*** our code base is being tested, which paves the way for undetected bugs.

See [here](https://github.com/pysal/spopt/pull/111#issuecomment-759569863) for my original comment.

PR #169, opened 2021-06-18
Increase test cases & coverage

jGaboardi:
This PR begins to address #112.

* blackens repo
* Clean up (some) docstrings
* Streamlines tests by reusing objects
* Fills out testing suite
  * [x] `azp`
    * basic AZP
    * ~~simulated annealing AZP~~ **code needs attention**
    * ~~Tabu AZP~~ **code needs attention**
      * ~~basic tabu~~ **code needs attention**
      * ~~reactive tabu~~  **code needs attention**
  * [x] `maxp`
    * `verbose == True`
    * complex cases (e.g., check move and simulated annealing)
    * one variable column for `attrs_name`
  * [x] `random_regions`
    * random regions
    * random region 
  * [x] `region_k_means`
    * reformatting and cleaning of test
    * add complex lattice test
  * [x] `skater`
    * test verbose
    * fix and test super verbose
    * add non-default parameters test
    * test islands (pass & fail)
  * [x] `spenc`
    * defaults, gamma, n_clusters
    * `spenc_lib`
  * [x] `ward` already 100% covered
    * reformatting and cleaning of test
  * [x] `util`
    * test solver checks

### DIFF
--- a/spopt/region/azp_util.py
+++ b/spopt/region/azp_util.py
@@ -152,7 +152,7 @@ class AllowMoveAZPMaxPRegions(AllowMoveStrategy):
 
         Parameters
         ----------
-        spatially_extensive_attr : numpy.ndarray`
+        spatially_extensive_attr : numpy.ndarray
             See corresponding argument in
             ``region.max_p_regions.heuristics.MaxPRegionsHeu.fit_from_scipy_sparse_matrix``.
         threshold : numbers.Real or numpy.ndarray
--- a/spopt/region/azp_util.py
+++ b/spopt/region/azp_util.py
@@ -7,24 +7,23 @@ import numpy as np
 
 
 class AllowMoveStrategy(abc.ABC):
-    def start_new_component(self, initial_labels, attr, objective_func,
-                            comp_idx):
+    def start_new_component(self, initial_labels, attr, objective_func, comp_idx):
         """
         This method should be called whenever a new connected component is
         clustered.
 
         Parameters
         ----------
-        initial_labels : :class:`numpy.ndarray`
+        initial_labels : numpy.ndarray
             The region labels of the areas in the currently considered
             connected component.
             Shape: number of areas in the currently considered component.
-        attr : :class:`numpy.ndarray`
+        attr : numpy.ndarray
             The areas' attributes.
             Shape: number of areas in the currently considered component.
-        objective_func : :class:`region.objective_function.ObjectiveFunction`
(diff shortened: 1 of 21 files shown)


jGaboardi:
I think this is about as good as we are going to get at the moment. The vast majority of code that is actually being used in regionalization models is now covered in test. The bigger issue is that there is a large amount of code that is simply not being used.

spopt/region/azp_util.py
```diff
@@ -152,35 +147,31 @@ class AllowMoveAZPMaxPRegions(AllowMoveStrategy):
...
         """
 
         Parameters
         ----------
-        spatially_extensive_attr : :class:`numpy.ndarray`, default: None
+        spatially_extensive_attr : numpy.ndarray
             See corresponding argument in
-            :meth:`region.max_p_regions.heuristics.MaxPRegionsHeu.fit_from_scipy_sparse_matrix`.
-        threshold : numbers.Real or :class:`numpy.ndarray`
+            ``region.max_p_regions.heuristics.MaxPRegionsHeu.fit_from_scipy_sparse_matrix``.
```
knaaptime, line 157: just a note to self, really, but we probably want a better description here that doesnt need to references region

jGaboardi: @knaaptime completely agree. When I started going through everything, I realized the docs will need a lot of love, so I focused on the testing coverage for his PR.

spopt/region/region_k_means.py
```diff
@@ -133,14 +133,14 @@ class RegionKMeansHeuristic(BaseSpOptHeuristicSolver):
 
     Attributes
     ----------
-    
-    labels_ : 
+
```
knaaptime, line 136: another note to self we'll wanna fill these in at some point

spopt/region/skater.py
```diff
@@ -318,10 +305,10 @@ def tqdm(noop, desc=""):
 
     def make_cut(self, in_node, out_node, score, MSF=None):
         """
-        make a cut on the MSF inplace, provided the in_node, out_node, MSF, and score. 
+        make a cut on the MSF inplace, provided the in_node, out_node, MSF, and score.
```
knaaptime, line 308: nbd, but we're switching docstring format in this function?

spopt/region/azp_util.py
knaaptime: lets do it in phases

Issue #187, opened 2021-08-11
The dissimilarity parameter should be able to accept a distance/dissimilarity matrix as an alternative to a callable distance function


Shruthi-Patil:
I'm currently working on spatial aggregation of energy systems data. Skater seems interesting for my problem and I would like to try it. 

However, the issue is that my data is 3-dimensional and has a complex structure. I compute the distance between each region pair and generate a matrix containing this info. 

It would be very helpful for my case and many other cases if the dissimilarity could be provided in the form of a pre-computed distance matrix, instead of a distance function.

jGaboardi:
@ljwolf @knaaptime @xf37 Would any of yall be able to look into this?

ljwolf:
Yeah, this should be easy to do. We would need to allow for "dissimilarity='precomputed'", as is common in other sklearn-style estimators. 

As it stands @Shruthi-Patil, you could try using `sklearn.cluster.AgglomerativeClustering()`, which currently supports both [precomputed affinity and connectivity constraints](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html)?

Shruthi-Patil:
@ljwolf I’m currently using the same. The results are okayish. That’s why we’re looking for alternatives. :)

ljwolf:
Interesting.... well I can take a look at customizing this very soon. I think the diff would be about 6 lines.

Shruthi-Patil:
Thanks. I look forward to it 😀

jGaboardi:
@Shruthi-Patil Do you still have any interest in seeing this move forward in #188? If not, we'll go ahead and close it out.

Shruthi-Patil:
@jGaboardi Thanks for the tweaks. However, we went with another algorithm. So please go ahead and close it.

PR #188, opened 2021-08-12
[WIP] allow for precomputed distance matrices

ljwolf:
This allows for precomputed distance matrices in the skater baseclass, `skater.SpanningForest`. For a usage example:

```python
import numpy
from libpysal import weights
from scipy.spatial import distance_matrix
from spopt.region.skater import SpanningForest

r = numpy.random.normal(size=(10,2))
D = distance_matrix(r,r, p=1) #l1 metric
w = weights.lat2W(5,2) # 5 by 2 lattice

SpanningForest(dissimilarity='precomputed').fit(5, w, D)
```

Now, caveat emptor: the semantics are a little different here @Shruti-Patil, since this converts the score into minimizing the sum of dissimilarities within the clusters, rather than minimizing the distance between features and the feature centroid of the cluster.

### DIFF
--- a/spopt/region/skater.py
+++ b/spopt/region/skater.py
@@ -168,7 +168,7 @@ class SpanningForest(object):
             if super_verbose:
                 print(self._trace[-1])
         if self.metric == "precomputed":
-            cutdata = dissim
+            cutdata = attribute_kernel
         else:
             cutdata = data
         while current_n_subtrees < n_clusters:  # while we don't have enough regions
@@ -256,11 +256,7 @@ class SpanningForest(object):
                 for l in range(n_subtrees)
             ]
         else:
-            part_scores = []
-            for l in range(n_subtrees):
-                label_data = data[labels == l]
-                values = label_data[label_data.nonzero()]
-                part_scores.append(self.reduction(values))
+            part_scores = [self.reduction(data[labels == l]) for l in range(n_subtrees)]
         return self.reduction(part_scores).item()
 
     def find_cut(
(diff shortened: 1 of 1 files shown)


ljwolf:
If you'd like to try it on your data, use `pip install git+https://github.com/ljwolf/spopt` and follow the example above.

jGaboardi:
Should we add a test for this or is it good without?

jGaboardi:
solution for #187

ljwolf:
I am not sure I want to merge this unless more people beyond @Shruti-Patil find it useful.... In abstract, it seems like a good idea, and I'm all for user power. But, empirically, I haven't seen good performance when minimizing the pre-computed dissimilarities.

I suppose the trick is that
- for the current implementation, decisions directly minimize the deviation *of the data* relative to its center (either median, mean, or any other user-supplied reduction). 
- In the precomputed case implemented here, we can only minimize the *dissimilarity* within the cluster. Elsewhere, there's no clear direction on whether this should be the *total* feature dissimilarity matrix, or the dissimilarity matrix *after filtering* by the possible joins (such that we're only considering the dissimilarity of "connected" observations). 
 
So, without further empirical work (on our end) to verify which is the right score, I don't want this to land.

jGaboardi:
> without further empirical work (on our end) to verify which is the right score, I don't want this to land.

This is reasonable.

jGaboardi:
@ljwolf Shall we go ahead and close this out as stale?

jGaboardi:
The OP [went in another direction](https://github.com/pysal/spopt/issues/187#issuecomment-1288487713) for a solution, so let's close.

PR #197, opened 2021-09-23
access package version in tools/gitcount

jGaboardi:
xref pysal/spaghetti#654, pysal/libpysal#434

In order to take advantage of using [`tools/gitcount.ipynb`](https://github.com/pysal/spopt/blob/master/tools/gitcount.ipynb) with the newly adopted `versioneer`schema, the version import cell must be split into two cells and changed from:

```python
# get __version__
f = "../{package}/__init__.py".format(package=package_name)
with open(f, "r") as initfile:
     exec(initfile.readline())
```

to:

```python
# get __version__
%run ../spopt/_version.py
```
```python
__version__ = get_versions()["version"]
```

### DIFF
--- a/tools/gitcount.ipynb
+++ b/tools/gitcount.ipynb
@@ -20,17 +20,14 @@
    "source": [
     "#  get date of last tag\n",
     "from subprocess import Popen, PIPE\n",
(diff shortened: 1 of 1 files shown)


gegen07 (approved):
LGTM.

martinfleis (approved):
The diff looks fine (I haven't run the notebook).

PR #205, opened 2021-11-03
install pulp for docs build

jGaboardi:
Seems that we need to install pulp for docs build. See failure [here](https://github.com/pysal/spopt/runs/4087183779?check_suite_focus=true).


jGaboardi:
Doc build locally without an issue after applying this fix. After this is merged we can do a `v0.2.1` release and simply copy over the `v0.2.0` Release Notes. Then we'll let JOSS know we are ready for reviewing to resume 🎉 .

gegen07 (approved):

PR #207, opened 2021-11-19
fix JOSS formatting

martinfleis:


jGaboardi (approved):
Thanks, @martinfleis !

PR #208, opened 2021-12-23
add 310 to CI matrix

jGaboardi:
The PR adds Python 3.10 to the GHA testing schema.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -15,11 +15,15 @@
      timeout-minutes: 30 
      strategy:
        matrix:
-         os: ['macos-latest', 'ubuntu-latest', 'windows-latest']
-         environment-file: [.ci/37.yaml, .ci/38.yaml, .ci/39.yaml, .ci/310.yaml]
+         os: [ubuntu-latest]
+         environment-file: [ci/37.yaml, ci/38.yaml, ci/39.yaml, ci/310.yaml]
          include:
-           - environment-file: .ci/310-DEV.yaml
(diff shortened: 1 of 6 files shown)


martinfleis:
CI change is fine. We now need to fix this line
https://github.com/pysal/spopt/blob/419ef748694c7b469e370c2f3bd5db8892b2662c/spopt/region/util.py#L68

I think that it needs to be `collections.abc.Iterable` now. That should work for 3.3+.

.github/workflows/unittests.yml
```diff
@@ -15,11 +15,15 @@
...
      strategy:
        matrix:
-         os: ['macos-latest', 'ubuntu-latest', 'windows-latest']
-         environment-file: [.ci/37.yaml, .ci/38.yaml, .ci/39.yaml]
+         os: [ubuntu-latest]
+         environment-file: [.ci/37.yaml, .ci/38.yaml, .ci/39.yaml, .ci/310.yaml]
          include:
-           - environment-file: .ci/39-DEV.yaml
+           - environment-file: .ci/310-DEV.yaml
              os: ubuntu-latest
```
martinfleis, line 22: With the current setup, you can remove this and add `310-DEV.yaml` to the list above on line 19.

jGaboardi: good catch!

martinfleis (approved):

gegen07 (approved):

martinfleis:
Thanks @jGaboardi!

PR #211, opened 2022-01-16
Xdist tests

jGaboardi:
This PR allows for multiprocessing in through `pytest`, which decreases runtime.

xref: pysal/spaghetti#661

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -12,7 +12,9 @@
    unittests:
      name: ${{ matrix.os }}, ${{ matrix.environment-file }}
      runs-on: ${{ matrix.os }}
-     timeout-minutes: 30 
+     timeout-minutes: 30
+     env:
+       RUN_TEST: pytest spopt --cov spopt -v -n auto --cov-config .coveragerc --cov-report xml --color yes --cov-append --cov-report term-missing
      strategy:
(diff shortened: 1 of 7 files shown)


.github/workflows/unittests.yml
```diff
@@ -12,7 +12,9 @@
    unittests:
      name: ${{ matrix.os }}, ${{ matrix.environment-file }}
      runs-on: ${{ matrix.os }}
-     timeout-minutes: 30 
+     timeout-minutes: 30
+     env:
+       RUN_TEST: pytest spopt --cov spopt -v -n auto --cov-config .coveragerc --cov-report xml --color yes --cov-append --cov-report term-missing
```
gegen07, line 17: Comparing to [spaghetti `.unittests.yml`](https://github.com/pysal/spaghetti/blob/main/.github/workflows/unittests.yml#L22), it's missing `--doctest-modules` directive. I think it will affect some tests, but it's better to handle it in another PR.

jGaboardi: `--doctest-modules` fails locally on 16 tests. This is because we have not been making sure examples within `docstrings` pass. As you said, better in another PR.

gegen07 (approved):

martinfleis (approved):

PR #221, opened 2022-03-10
mclp and lscp jupyter notebooks using osrm routing

TimMcCauley:
Greetings @martinfleis @jGaboardi - I refactored the notebook a little and added some comments. Please let me know what you think. Happy to work in feedback if you have ideas or comments. I'd go for LSCP up next (please choose your city :) ) !


jGaboardi:
Thanks so much for this @TimMcCauley! Here are some comments:
* Since `folium` and `routingpy` are needed here, but not for `spopt` in general, we may want to have a statement about that in the notebook description.
* It looks like the notebook starts from cell 209. Can you rerun so it starts from cell 1?
* There are some long lines of code in there. We should be trying to stick to [PEP8](https://peps.python.org/pep-0008/#maximum-line-length) guidelines as much as possible, but we go to a line length of 90 frequently. You can reformat manually or use [`black`](https://black.readthedocs.io/en/stable/) within the notebook with [`drillan/jupyter-black`](https://github.com/drillan/jupyter-black).

TimMcCauley:
Thank you @jGaboardi - I took care of your points and pushed an update.

gegen07 (commented):
@TimMcCauley I didn't know about the routingpy package. That's a great one, nice addition.

I have some points:
- I think the map is kinda small, mainly when we got the results. We could improve increasing the zoom_start to 15 and change the start location to `37.748273, -122.410558`. 
- The mission district geojson structure is hardcoded. We could change it by using Nominatim API, it's just a request. I'm also fine if @jGaboardi and @martinfleis think that hardcoded polygon is ok in that case.
- The polygon that you are using is a merge of two districts I think, Bernal Heights and Mission District. It would be good to remind that in the description or even in the comments that you did above the polygon definition line.

Awesome notebook!!! 🎉

TimMcCauley:
@gegen07 thank you - I updated it & it looks better now. I'd keep the nominatim request out for this NB if it's ok but I am currently writing one for Toronto and LSCP and I will make sure to include it there.

TimMcCauley:
I added the LSCP notebook in this PR, too. You can find it in `notebooks/lscp_gis.ipynb` or check it out [here](https://nbviewer.org/github/gis-ops/spopt/blob/main/notebooks/lscp_gis.ipynb) - let me know what you think. I made use of the Overpass API this time @gegen07 - let me know what you think (the use case is a little exotic but why not! ;-) )

jGaboardi:
Very nice with the LSCP notebook. One small typo I saw between cells 2 and 3 where **Solving the MCLP** should be **Solving the LSCP**.

TimMcCauley:
> Very nice with the LSCP notebook. One small typo I saw between cells 2 and 3 where **Solving the MCLP** should be **Solving the LSCP**.

Oh, right. Fixed that now!

jGaboardi (approved):

gegen07 (approved):

jGaboardi:
Think this is good for merging. Thanks for the contribution @TimMcCauley 🎉 !

PR #231, opened 2022-04-17
Add predefined locations to coverage models

gegen07:
Based on ideas raised in discussion #218. 
I'm trying to implement @jGaboardi idea, for instance.


jGaboardi:
This is a very nice start @gegen07!

jGaboardi:
I just realized that we are not assigning decision variable labels for plotting, but only using the `range` (see [`plot_results`](https://github.com/pysal/spopt/blob/72059b2c6d0c06025905277c83b3bb8e8600fe83/notebooks/mclp.ipynb) in the MCLP notebook). By doing this we have the same labels for the solution facilities in "cost matrix" and "geodataframe" examples, though the selection set is different. In order to ensure comparable results (and confirm the preselected facilities are being adhered to) we need to label the facilities from a static index (e.g. decision variable label from the `pulp` model) and also ensure a 1:1 color scheme (e.g.

gegen07:
In the MCLP notebook we have different labels in the legend. We are not getting the label from the pulp model to create the facility-client matrix. When do you say create a static index you mean to get the labels from pulp, like another method to get these labels, or just assign a 1:1 dictionary in the `plot_results` method using the indexes of the facility-client matrix?

jGaboardi:
> In the MCLP notebook we have different labels in the legend.
> We are not getting the label from the pulp model to create the facility-client matrix. When do you say create a static index you mean to get the labels from pulp, like another method to get these labels, or just assign a 1:1 dictionary in the `plot_results` method using the indexes of the facility-client matrix?

ah! You are right about that, I completely misread the legend. My fault. Though it is still good to have a defined color map so that a specific color is tied to a specific facility. Then the two plots in the notebook are easier to compare.

jGaboardi (approved):
This is unbelievably great work @gegen07. Let's also try to get a reviews from Qunshan and Levi if they have time. But it looks great to me.

jGaboardi:
I think this is probably good to go and we can go ahead and merge. 🎉

qszhao:
Thanks @jGaboardi and @gegen07! Great to have this question sorted out!

Issue #241, opened 2022-06-05
Consider documenting units for the `max_coverage` parameter of MCLP/LSCP.from_geodataframe


apulverizer:
Noticed this as part of https://github.com/openjournals/joss-reviews/issues/3330

It's a bit unclear what the distance unit actually is. I assume it's in the units of the coordinate system that the geodataframes are in. From a usability perspective, it'd be nice to at least document that. Even better, would be to support specifying the units and internally converting it. 

Something like:

```python
mclp_from_geodataframe = MCLP.from_geodataframe(
    clients_snapped, 
    facilities_snapped, 
    "geometry", 
    "geometry", 
    "demand", 
    100,
    service_radius_units="miles",
    p_facilities=P_FACILITIES,
    distance_metric="euclidean",
)
```

jGaboardi:
@apulverizer This is a nice suggestion, and one we have considered before. However, as with other PySAL submodules, we expect the user to have performed preprocessing and to know the projection/coordinate system. Therefore, the user must specify only distance unit itself. Considering the naming [`max_coverage` was recently changed to `service_radius`](https://github.com/pysal/spopt/commit/7bcb35374885b4812dc7845194fd99ee5f0540b4) in order to more accurately reflect facility location terminology.

Issue #240, opened 2022-06-05
Consider documenting class properties


apulverizer:
While I was reviewing this package as part of https://github.com/openjournals/joss-reviews/issues/3330 I noticed that quite a few of the tutorials use undocumented properties of the models for plotting and/or statistics. 

For example:
https://github.com/pysal/spopt/blob/360ea6cbfdd81655874811f50dc26db865ad533f/spopt/locate/coverage.py#L277

is used in the sample MCLP notebook: https://github.com/pysal/spopt/blob/main/notebooks/mclp.ipynb?short_path=d3109c7#L944

I couldn't find any doc on this property in the [API ref](https://pysal.org/spopt/generated/spopt.locate.coverage.MCLP.html#spopt.locate.coverage.MCLP). From a coding perspective I expected `facility_client_array` to return the 2x2 matrix, not `None`. It's odd I have to call that method and then access an undocumented property to use it...

jGaboardi:
Resolved via #243.

PR #243, opened 2022-06-06
[WIP] JOSS Enhancements

gegen07:
- [ ] Add results parameters to solve method
- [ ] Document `fac2cli`, `cli2fac`, `n_cli_uncov` properties

### DIFF
--- a/spopt/locate/p_center.py
+++ b/spopt/locate/p_center.py
@@ -280,7 +280,7 @@ class PCenter(LocateSolver, BaseOutputMixin):
 
             self.fac2cli.append(array_cli)
 
-    def solve(self, solver: pulp.LpSolver):
+    def solve(self, solver: pulp.LpSolver, results: bool = True):
         """
         Solve the PCenter model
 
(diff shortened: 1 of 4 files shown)


spopt/locate/p_center.py
```diff
@@ -280,19 +280,28 @@ def facility_client_array(self) -> None:
...
         -------
         PCenter object
         """
         self.problem.solve(solver)
         self.check_status()
+
+        if results:
+            self.facility_client_array()
+            self.client_facility_array()
+            self.uncovered_clients()
```
jGaboardi, line 305: With $p$-center and $p$-median there will be no uncovered clients. Same goes for Set Cover. The only model with uncovered clients will be the Max Cover.

jGaboardi:
Addressing #240 and #241 raising from openjournals/joss-reviews#3330.

martinfleis:
Can you rename the PR using something that describes the changes? This will not be very helpful later when tracking changes.

spopt/locate/p_center.py
jGaboardi: Great job, @gegen07

gegen07:
@jGaboardi I think it's ready for review!

jGaboardi (approved):
Very nicely done @gegen07.

Issue #213, opened 2022-01-23
K means heuristic test failure


jGaboardi:
Need to investigate [this failure](https://github.com/pysal/spopt/runs/4913930765?check_suite_focus=true#step:6:234) for the `region` *k*-means heuristic in Python 3.10.

jGaboardi:
The test does not fail locally with Python 3.10.2.

jGaboardi:
This appears to have worked itself out somewhere along the way. All tests are green again.

jGaboardi:
Seems to be failing again for Python 3.10 on Windows, [but only sometimes](https://github.com/pysal/spopt/runs/6853689627?check_suite_focus=true).

jGaboardi:
This is causing headaches again. 

https://github.com/pysal/spopt/runs/7520297888?check_suite_focus=true

jGaboardi:
This failure is really messy with development efforts. Let's mark the [`test_region_k_means_heuristic_synth_small`](https://github.com/pysal/spopt/blob/a17c8da5132cd8b77db3738af377adeb1d01abfd/spopt/tests/test_region_k_means.py#L36-L39) with an [`xfail`](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) until we work out the situation.

PR #250, opened 2022-06-13
clean out code block and add preexisting facility text

jGaboardi:
@gegen07 This PR cleans out the MCLP code block and adds in a couple sentences about the pre-existing facilities functionality.

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -19,13 +19,13 @@ authors:
     affiliation: 1
   - name: Ran Wei
     affiliation: 1
-  - name: Levi Wolf
+  - name: Levi J. Wolf
     orcid: 0000-0003-0274-599X
     affiliation: 6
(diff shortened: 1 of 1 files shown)


jGaboardi:
xref openjournals/joss-reviews#3330

gegen07 (commented):
@jGaboardi LGTM! Do you know why kmeans heuristic test failed?

jGaboardi:
> @jGaboardi LGTM! Do you know why kmeans heuristic test failed?

It passes sometimes and fails sometimes. No real idea what's going with it other than maybe something in `scipy`. I had an issue open for it (see #213), but all had been passing again until the past few weeks. We don't have to worry about it for now, but if the failure keeps going for a couple more weeks we can reopen #213 and take a deeper look.

Issue #258, opened 2022-06-22
Locate notebooks are giving errors


gegen07:
With API updates #241 and #242, `max_coverage` parameter changed to `service_radius` and only MCLP model `has uncovered_clients` attribute. Because of these, `plot_results` is raising an `AttributeError`.

Issue #253, opened 2022-06-13
Example Notebook: LSCP 


erinrolson:
![Screen Shot 2022-06-13 at 12 16 46 PM](https://user-images.githubusercontent.com/39073850/173398822-5f024a76-bb55-4fa4-9e85-10ed616938f8.png)

I was going through the [LSCP notebook](https://pysal.org/spopt/notebooks/lscp.html) in detail and am confused by section 19 "LSCP built from cost matrix". lscp_from_cost_matrix.facility_client_array() returns a NoneType object for me and I don't see anywhere in the notebook where results from this line are ever used. If I comment the line out the results still plot.

jGaboardi:
Correct. This is due to [functionality](https://github.com/pysal/spopt/blob/abc3a764beec1ed6fcd02f4788b905076b7ef6a7/spopt/locate/coverage.py#L311-L312) that @gegen07 included last week that call the `{client_facility/facility_client}_array()` methods internally when solving. Moreover, nothing is returned when calling the `{client_facility/facility_client}_array()` methods because the [`cli2fac`](https://github.com/pysal/spopt/blob/abc3a764beec1ed6fcd02f4788b905076b7ef6a7/spopt/locate/base.py#L72) and [`fac2cli`](https://github.com/pysal/spopt/blob/abc3a764beec1ed6fcd02f4788b905076b7ef6a7/spopt/locate/coverage.py#L280) attributes are being set.

jGaboardi:
@gegen07 This is something we should probably update in the notebooks.

PR #261, opened 2022-07-16
Locate Notebooks with API inconsistent usage

gegen07:
Solves #258.

It also changed the service radius of `lscp_gis.ipynb`. Using 2 minutes it was giving an infeasible solution. I checked the sum over clients and discovered there is a client with a minimum range of 147.9 seconds to a facility, so I just updated the model radius to 150 seconds.


gegen07:
xref #253

jGaboardi (approved):

jGaboardi:
Nice work!

PR #264, opened 2022-07-26
mark k_means_heuristic_synth_small test with xfail 

jGaboardi:
This PR provides a temporary solution to #213 until we figure out what to do more permanently.


jGaboardi:
xref #259

ljwolf:
Looks fine to me!

PR #273, opened 2022-09-08
Facility capacity for LSCP

erinrolson:
Implements facility capacity constraint for solving LSCP for GSoC '22


notebooks/lscp_capacity.ipynb
```diff
@@ -0,0 +1,1129 @@
...
+    "& & d_{ij} & \\small = & \\textrm{shortest distance or travel time between nodes} \\quad i \\quad \\textrm{and} \\quad j \\\\\n",
+    "& & N_i & \\small = & \\{j | d_{ij} < S\\} \\\\\n",
+    "& & x_j & \\small = & \\begin{cases} \n",
+    "          1, \\text{if a facility is located at node} \\quad j\\\\\n",
+    "          0, \\text{otherwise} \\\\\n",
+    "       \\end{cases} \\end{array}$\n",
+    "       \n",
+    "_This excerpt above was quoted from Church L., Murray, A. (2018)_\n",
+    "\n",
+    "This tutorial solves LSCP using `spopt.locate.coverage.LSCP` instance that depends on a array 2D representing the costs between facilities candidate sites and demand points. For that it uses a lattice 10x10 with simulated points to calculate the costs."
```
erinrolson, line 37: Update intro with new description and reference/display new capacity formulation

jGaboardi:
@qszhao

jGaboardi:
@ljwolf conflicts were minor and now are resolved.

spopt/locate/base.py
```diff
@@ -476,6 +484,82 @@ def add_predefined_facility_constraint(
...
+        None
+        """
+        if hasattr(obj, "fac_vars") and hasattr(obj, "cli_assgn_vars"):
+            cli_assn_vars = getattr(obj, "cli_assgn_vars")
+
+            for i in range_client:
+                model += pulp.lpSum([cli_assn_vars[i][j] for j in range_facility]) == 1
+        else:
+            raise AttributeError(
+                "before setting constraints must set facility variable and demand quantity variable"  # might want to update this message later
```
ljwolf, line 560:
```suggestion
"The facility variable and demand quantity variable most both be set in order to add a client demand satisfaction constraint."
```

ljwolf (changes requested):
I have one suggestion on the error warning, and would also suggest a bit more narrative in the example notebook. But, I'll need to commit that separately from this review.

ljwolf:
OK, I've updated the notebook wording. I will merge once the tests pass.

jGaboardi (approved):

ljwolf:
Hmm, I missed that for sure. I think that the notebook serves as a user test, and the existing option passing is tested already. What's not tested is the more finicky stuff, like argument typing and things... I think it's OK here because the functionality is completely checked by the notebook, but I wouldn't want this to justify future code without tests!

jGaboardi:
> I think that the notebook serves as a user test, and the existing option passing is tested already. What's not tested is the more finicky stuff, like argument typing and things... I think it's OK here because the functionality is completely checked by the notebook, but I wouldn't want this to justify future code without tests!

My thoughts exactly. I will look into adding to @erinrolson's excellent contribution by doing a thorough doc review, adding tests, and probably adding some spice to the notebook. Hopefully I can get to this in the next couple weeks so we can include in the coming meta release.

PR #294, opened 2022-10-30
Region-k-means seed

### LINKED ISSUE #213: K means heuristic test failure

jGaboardi:
Resolves #213.

After more digging I found that we were not providing the functionality of actually passing in a seed value to the [`base._seeds()`](https://github.com/pysal/spopt/blob/main/spopt/region/base.py#L223-L244) function. Therefore, the resultant `numpy.random.choice()` function within could never be reproduced and was screwing tests up.

This PR also fills in all docstrings in `base.py`, cleans up several raised error logics, and adds several more test cases.

### DIFF
--- a/spopt/region/base.py
+++ b/spopt/region/base.py
@@ -37,15 +37,12 @@ def move_ok(area, source, destination, g, w):
     Parameters
     ----------
 
-    area :
-        ...
-
-    source :
-        ...
-
-    destination :
-        ...
-
+    area : int, numpy.int64
+        Area label.
+    source : numpy.array
+        Source region labels.
+    destination : list
+        destination region labels.
     g : networkx.Graph
         NetworkX representation of ``w``.
     w : libpysal.weights.W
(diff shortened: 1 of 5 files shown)


jGaboardi:
@sjsrey @knaaptime gently pining for review (when time permits).

spopt/region/base.py
```diff
@@ -7,52 +7,23 @@
 from scipy.spatial import KDTree
 
 
-class RegionMixin(object):
```
sjsrey, line 10: Is `RegionMixin` going away?

sjsrey (commented):
The seed stuff looks good. Just one question related to RegionMixin that I left.

spopt/region/base.py
jGaboardi: It's not used anywhere within `spopt` and it seemed like it was legacy from the days of `pysal/region`. Should we keep it? We can absolutely keep it if that would be prudent.

sjsrey: I think the intention was to have a common set of methods in all the models to enforce standardization (`model.solve`). So I think we should have a discussion of whether that is still desired - if it is then we can go back and refactor existing classes to adopt that, if not, then we can merge this and do away with the Mixin.

jGaboardi: How about I add it back in here now and we can open an issue on whether to get rid of it the future? Probably better to have time for discussion, as you say.

jGaboardi:
converted #297 to a discussion (#298)

sjsrey (approved):
Thanks for spotting the lack of use of RegionMixin

Issue #280, opened 2022-10-14
short-to-mid term plan for improving docs, structure, and formatting


jGaboardi:
Following #279, we should focus back on:
* exposing & documenting current functionality
  * [ ] #101 
  * [ ] #170 
  * [ ] #172
  * [ ] #173
  * [ ] #269
* rerun all notebooks and the binder to ensure functional
  * [ ] #276
  * [ ] #278 
* automate formatting standards, etc.
  * [ ] #210 
  * add linting to CI and `pre-commit`
    * `black`, `isort`, `flake8`
    * see [`momepy`](https://github.com/pysal/momepy) as an example

Issue #173, opened 2021-06-25
fill and and standardize docstrings


jGaboardi:
Docstrings within `region` models are in several formats and there are quite a few missing argument explanations and example uses. See also @knaaptime's comments [here](https://github.com/pysal/spopt/pull/169#pullrequestreview-692958180), [here](https://github.com/pysal/spopt/pull/169#pullrequestreview-692962310), and [here](https://github.com/pysal/spopt/pull/169#pullrequestreview-692963966).

This is a running list of tasks. Update as needed.
* [ ] `azp`
  * ...
* [ ] `maxp`
  * ...
* [ ] `random_regions`
  * ...
* [ ] `region_k_means`
  * ...
* [ ] `skater`
  * ...
* [ ] `spenc`
  * ...
* [ ] `ward`
  * ...

PR #296, opened 2022-10-31
New attributes & functionality for `locate` models

### LINKED ISSUE #267: pre-defined facilities for all location models?
It looks like the functionality to stipulate pre-defined facilities never made it into the [`PCenter`](https://github.com/pysal/spopt/blob/main/spopt/locate/p_center.py) or [`PMedian`](https://github.com/pysal/spopt/blob/main/spopt/locate/p_median.py) classes. I had sworn we got that in, but it must have been my oversight.

### LINKED ISSUE #282: maybe set solver as attribute in LSCPB?
[`LSCPB.from_cost_matrix()`](https://github.com/pysal/spopt/blob/main/spopt/locate/coverage.py#L365) and [`LSCPB.from_geodataframe()`](https://github.com/pysal/spopt/blob/main/spopt/locate/coverage.py#L474) take `solver` as a positional argument in order to solve the LSCP internally. Might it be more suitable to then set `solver` as an attribute for `LSCPB` so that it does not need to be passed into [`LSCPB.solve()`](https://github.com/pysal/spopt/blob/336f235843ec86bf9ad1535f6624c15ae9ff298f/spopt/locate/coverage.py#L606)? We probably don't want to be solving the LSCP internally and LSCPB with different solvers. Only allowing `solver` to be passed in one time would alleviate this.

This pattern would be different than the other models, but I think it's worth it.

Thoughts? @erinrolson @gegen07 @ljwolf @qszhao

### LINKED ISSUE #283: get_backup_percentage method for LSCP-B
Similar to [`get_percentage`](https://github.com/pysal/spopt/blob/main/spopt/locate/base.py#L114) in MCLP and [`get_mean_distance`](https://github.com/pysal/spopt/blob/main/spopt/locate/base.py#L131) in PMedian, it would be nice to have a `Mixin` class + method for calculating the percentage of clients with backup in the `LSCPB` class. Possibly something like:

```python
class BackupPercentageMixin:
    def get_backup_percentage(self):
        self.backup_percentage = self.problem.objective.value() / len(self.cli_vars)

```

### LINKED ISSUE #285: remove BaseOutputMixin from PDispersion
Currently [`PDispersion`](https://github.com/pysal/spopt/blob/main/spopt/locate/p_dispersion.py#L16) is initialized with the `LocateSolver` and [`BaseOutputMixin`](https://github.com/pysal/spopt/blob/main/spopt/locate/base.py#L58) classes. However, the $p$-dispersion problem does not take clients into account so we should not be using the `BaseOutputMixin` there. The only method within `BaseOutputMixin` is [`client_facility_array()`](https://github.com/pysal/spopt/blob/main/spopt/locate/base.py#L63).

### LINKED ISSUE #286: calculate `get_mean_distance()` within ``PMedian.solve(results=True)``
We should move the calculation of [`get_mean_distance()`](https://github.com/pysal/spopt/blob/336f235843ec86bf9ad1535f6624c15ae9ff298f/spopt/locate/base.py#L131) within [`PMedian.solve(results=True)`](https://github.com/pysal/spopt/blob/336f235843ec86bf9ad1535f6624c15ae9ff298f/spopt/locate/p_median.py#L331). This would necessitate adding [`weights_sum`](https://github.com/pysal/spopt/blob/main/spopt/locate/p_median.py#L160) as an attribute to `PMedian` instances, but that is also probably not a bad idea in and of itself.

### LINKED ISSUE #287: calculate `get_percentage()` within ``MCLP.solve(results=True)``
We should move the calculation of [`get_percentage()`](https://github.com/pysal/spopt/blob/main/spopt/locate/base.py#L114) within [`MCLP.solve(results=True)`](https://github.com/pysal/spopt/blob/336f235843ec86bf9ad1535f6624c15ae9ff298f/spopt/locate/coverage.py#L956).

Also, let's:
* rename the calculated attribute from [`percentage`](https://github.com/pysal/spopt/blob/main/spopt/locate/base.py#L119) to `percentage_covered`/`perccov`/`perc_cov`.
* calculate the actual percentage: `(1 - (self.n_cli_uncov / self.aij.shape[0])) * 100.`

jGaboardi:
This PR:
* resolves #267 
* resolves #282
* resolves #283 
* resolves #285 
* resolves #286 
* resolves #287 
* reruns notebooks to demonstrate new functionality & attributes
* starts a cleans up and standardization of docstrings (this will continue in future PRs)
* starts a cleanup and debugging of doctests (this will continue in future PRs)
  * all doctests in `locate` now pass, but ones `region` need much work (xref #173, #280)


jGaboardi:
xref #295 needs attention in the short term.

spopt/tests/test_locate.py
```diff
@@ -375,26 +399,65 @@ def test_pmedian_client_facility_array_from_geodataframe(self):
...
 
+    def test_p_center_from_cost_matrix_no_results(self):
+        p_center = PCenter.from_cost_matrix(self.cost_matrix, p_facilities=4)
+        result = p_center.solve(pulp.PULP_CBC_CMD(msg=False), results=False)
+        assert isinstance(result, PCenter)
+
+        with pytest.raises(AttributeError):
+            result.cli2fac
+        with pytest.raises(AttributeError):
+            result.fac2clif
```
gegen07, line 453: Typo here. Instead `result.fac2clif` should be `result.fac2cli`. The same for lines 330, 196, 81.

spopt/locate/coverage.py
```diff
@@ -371,59 +461,90 @@ def from_cost_matrix(
...
+
+        Get the percentage of clients covered by more than one facility.
+
+        >>> round(lscpb_from_cost_matrix.backup_perc, 3)
+        88.0
+
+        88% of clients are covered by more than 1 facility
 
         """
         if predefined_facilities_arr is not None:
```
gegen07, line 550: Should predefined facilities be equal for LSCP and LSCPB when solving LSCPB? Does that make sense give an option to pass another column to distinguish predefined sites between these models?

spopt/tests/test_lscpb.py
```diff
@@ -62,19 +62,32 @@ def test_lscpb_from_cost_matrix(self):
...
+        lscpb = LSCPB.from_cost_matrix(
+            self.cost_matrix, 10, pulp.PULP_CBC_CMD(msg=False)
+        )
+        result = lscpb.solve(results=False)
+        assert isinstance(result, LSCPB)
+
+        with pytest.raises(AttributeError):
+            result.cli2fac
+        with pytest.raises(AttributeError):
+            result.fac2clif
```
gegen07, line 79: Typo here. Instead result.fac2clif should be result.fac2cli.

gegen07 (changes requested):
hey @jGaboardi! Wow, nice improvements! I made minor comments.

spopt/locate/coverage.py
jGaboardi: This is an interesting question – I think they should be equal. Since the "predefined"ness of the LSCPB is derived from the LSCP solution, those facilities that are required to be in the LSCPB solution must also be passed into the LSCP.

Can you think of a scenario where this would not be the case? I am open to different ideas/opinions.

jGaboardi: @qszhao & @ljwolf, how do you feel about this point that @gegen07 brought up here?

gegen07: I couldn't think of a case that this occurs. Hmm, but I imagined that if we solve the LSCP with predefined facilities we will already choose them in the results, so these sites will not interfere with LSCPB solving. So, we can disconsider the predefined locations for LSCPB, if we assume they are the same. What do you think @jGaboardi?

jGaboardi: But we still need to pass the predefined facilities into the LSCPB instance in order to ensure they are part of the LSCPB solution. This is because the standard LSCPB takes into consideration the objective of the LSCP (*number* of facilities, but not *which* facilities). So I think the correct answer is to leave it as is.

gegen07: OK! I think we can go on and merge this PR.

gegen07 (approved):

PR #309, opened 2022-11-26
workaround for cascading references in docs

### LINKED ISSUE #303: weird cascading references in docs
In the *References* section of the `spopt` doc site, when a reference occupies a single line the subsequent reference is indented. This leads to a cascading effect that is only broken when a reference is more than one line long. I have not been able to determine why this is happening or a solution for it, but my hypothesis is that it has something to do with a setting in [`pysal-styles.css`](https://github.com/pysal/spopt/blob/main/docs/_static/pysal-styles.css). An example of this can be seen in the [*References* section](https://pysal.org/spopt/references.html) of the `spopt` doc site, and from the screenshot below:

Screenshot 2022-11-12 at 11 51 33 AM

Another example can be seen in the [`spaghetti` docs](https://pysal.org/spaghetti/references.html). Any assistance on debugging this would be greatly appreciated. I have tried searching SO and the like, but I don't seem to be using the correct search terms...

@weikang9009 @knaaptime @sjsrey

jGaboardi:
This PR
* provides a workaround for #303


martinfleis:
This is not really solving the cause right? Just the result. Do we know where the original issue comes from apart from versions? It may as well be a bug in sphinx that should be resolved there, not by a CSS fix here.

jGaboardi:
> This is not really solving the cause right? Just the result. Do we know where the original issue comes from apart from versions? It may as well be a bug in sphinx that should be resolved there, not by a CSS fix here.

Correct.

martinfleis:
Okay. Then it should be reported to https://github.com/ryan-roemer/sphinx-bootstrap-theme/issues and fixed there imho. We can still make our custom patch but that should not be a long term solution.

jGaboardi:
> Okay. Then it should be reported to https://github.com/ryan-roemer/sphinx-bootstrap-theme/issues and fixed there imho. We can still make our custom patch but that should not be a long term solution.

Yeah, you're right. I had a small reproducible example somewhere, but can't find it now. After I get that either found or created again, I will report to `sphinx-bootstrap-theme` and merge this for a temp fix.

PR #316, opened 2022-12-18
Fix CLSCP-SO bugs and add tests

### LINKED ISSUE #314: [BUG] service radius threshold not respected in CLSCP-SO
xref #273

In the Capacitated Location Set Cover Problem – System Optimal (CLSCP-SO), the service radius is not being considered when [generating the client demand satisfaction constraints](https://github.com/pysal/spopt/blob/10535e6f9f59527cd5348766c8f568706c236395/spopt/locate/base.py#L557). We need to be passing the masked cost matrix into [`add_client_demand_satisfaction_constraint()`](https://github.com/pysal/spopt/blob/10535e6f9f59527cd5348766c8f568706c236395/spopt/locate/coverage.py#L225-L226) and then adding that to the constraint set, as in [`add_set_covering_constraint()`](https://github.com/pysal/spopt/blob/10535e6f9f59527cd5348766c8f568706c236395/spopt/locate/base.py#L354). This is demonstrated below from [the notebook](https://github.com/pysal/spopt/blob/main/notebooks/lscp_capacity.ipynb) where a valid optimal solution (all client locations are covered) is found  in scenario 1 (original), but only 3 client locations are covered when capacities are switched (scenario 2), despite the model being solved as optimal.
Screenshot 2022-12-17 at 8 29 32 PM

Screenshot 2022-12-17 at 8 29 45 PM

### LINKED ISSUE #315: raise if demand > capacity in CLSCP-SO
Currently there is no check that capacity >= demand when formulating the [CLSCP-SO](https://github.com/pysal/spopt/blob/main/spopt/locate/coverage.py#L190). We need a check there that raises a `ValueError` if demand >= capacity.

xref #314

jGaboardi:
This PR:
* resolves #314 
* resolves #315 
* add basic testing suite for CLSCP-SO
* reruns notebook


spopt/locate/coverage.py
```diff
@@ -208,22 +208,30 @@ def from_cost_matrix(
...
-                lscp.aij,
                 facility_capacity_arr,
                 demand_quantity_arr,
                 r_fac,
                 r_cli,
             )
 
             FacilityModelBuilder.add_client_demand_satisfaction_constraint(
-                lscp, lscp.problem, r_cli, r_fac
+                lscp, lscp.problem, lscp.aij, r_cli, r_fac
```
sjsrey, line 234: is it redundant to pass in lscp, lscp.problem, lscp.aij as separate args, when the latter two are attributes/members of the former?

sjsrey (commented):
Looks good. One question for my education more than anything.

spopt/locate/coverage.py
jGaboardi: This is a very good point. I'd say let's open an issue to standard are streamline variable & constraint creation. Then we can merge this PR in and work on that after?

jGaboardi: xref #317

jGaboardi: So I'll go ahead with the merge and get to work on #317.

Issue #331, opened 2023-01-06
`spopt==0.5.0` release


jGaboardi:
We need to cut the next release of `spopt` in the coming week, probably end of next week at the latest. This next release will be `0.5.0`. I would like the following issues addressed and PRs merged before that, if at all possible.
* ***highly desired***
  * [x] #321
     * The PR is up and ready for review – #333 
     * The PR is up and ready for review – #335 
  *  [x] #320
     * The PR is up and ready for review – #329
* ***stretch***
  * [x] #328 
    * The PR is up and ready for review – #336
  * [x] #332

jGaboardi:
This is now ready to go! May start with release candidates to make sure the docs build and release workflow still function as expected.

jGaboardi:
[`spopt==0.5.0rc1`](https://pysal.org/spopt/index.html#) is live. Will wait a couple days to let it filter downstream through `conda-forge` before going forward with `spopt==0.5.0`.

knaaptime:
is that as simple as including a tag with the rc suffix?

jGaboardi:
Yep

```
$ git tag -a v#.#.#rc#
$ git push upstream v#.#.#rc#
```

PR #336, opened 2023-01-10
[BUG] bounds for client assignment variables

### LINKED ISSUE #328: [BUG] client assignment variable bounds for CLSCP-SO
In the [`add_client_assign_integer_variable`](https://github.com/pysal/spopt/blob/172b356c444bb1af82d8eb3aefb291976fb3bd61/spopt/locate/base.py#L235) factory, the variable lower and upper bounds are hard-coded to 0 and 1, respectively. This is correct for the $p$-median and $p$-center models. However in the CLSCP-SO formulation, there is no upper bound on the assignment variables (see equation [7.28] of [Church & Murry 2018](https://link.springer.com/book/10.1007/978-3-319-99846-6)):
$$z_{ij} \geq 0$$
Moreover, in the CLSCP-SO the client assignment variables are not constrained to be integers, which we already [account for by setting them as continuous](https://github.com/pysal/spopt/blob/172b356c444bb1af82d8eb3aefb291976fb3bd61/spopt/locate/coverage.py#L221-L222).

So we need to *either*:
1. rename the method and add keywords for lower/upper bounds
   * `add_client_assign_integer_variable` -> `add_client_assign_variable`
2. create a new method
   * `add_client_assign_continuous_variable`

Choice (1.) is the most elegant and maintainable solution, IMHO.

jGaboardi:
This PR resolves #328. Following merge we can cut release `v0.5.0` 🎉 

xref #331

### DIFF
--- a/notebooks/lscp_capacity.ipynb
+++ b/notebooks/lscp_capacity.ipynb
@@ -60,8 +60,8 @@
    "execution_count": 1,
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2023-01-10T18:19:53.041361Z",
-     "start_time": "2023-01-10T18:19:53.007862Z"
+     "end_time": "2023-01-10T23:04:25.556826Z",
+     "start_time": "2023-01-10T23:04:25.511104Z"
     }
    },
    "outputs": [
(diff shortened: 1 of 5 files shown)


gegen07 (approved):
Nice catch!

PR #344, opened 2023-01-13
Add `requirements_notebooks.txt` to MANIFEST

### LINKED ISSUE #343: The requirementes notebook is missing from the MANIFEST.in
Ping @jGaboardi 

See https://github.com/pysal/spopt/blob/4500eb9c93e8d501dd72ac6eb940d1125c5adee8/MANIFEST.in#L1

jGaboardi:
This PR resolves #343.

xref https://github.com/conda-forge/spopt-feedstock/pull/12#issuecomment-1382356827

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,3 @@
-include LICENSE.txt MANIFEST.in requirements.txt requirements_docs.txt requirements_dev.txt requirements_tests.txt
+include LICENSE.txt MANIFEST.in requirements.txt requirements_docs.txt requirements_dev.txt requirements_tests.txt requirements_notebooks.txt
 include versioneer.py
 include spopt/_version.py


martinfleis:
not a huge fan of this solution of loading all reps from n files on setup (pyproject.toml would be certainly better) but it is fine now. 

long-term, none of these should be in manifest as they're not needed in the wheel.

jGaboardi:
> not a huge fan of this solution of loading all reps from n files on setup (pyproject.toml would be certainly better) but it is fine now.
> 
> long-term, none of these should be in manifest as they're not needed in the wheel.

FWIW, I agree. After we work the kinks out of using `pyproject.toml` exclusively (it's causing a headache over in [`spaghetti`](https://github.com/pysal/spaghetti/issues/701)) then I'll start pushing for adoption across the PySAL ecosystem.

PR #351, opened 2023-02-13
Update regionalisation demo in README

paddyroddy:
Current first example doesn't work


jGaboardi (approved):

jGaboardi:
Good catch @paddyroddy Great catch and thanks for the PR!

Issue #346, opened 2023-02-02
failing `region_k_means` test with bleeding edge `libpysal`


jGaboardi:
This seems to have started [failing](https://github.com/pysal/spopt/actions/runs/4070255241/jobs/7010902891#step:8:346) with https://github.com/pysal/libpysal/pull/511. However, adding either `sort_joins={True, False}` to [this line](https://github.com/pysal/spopt/blob/main/spopt/region/region_k_means.py#L61) results in the same failure. Perhaps I am missing an obvious detail?

```
[gw1] darwin -- Python 3.11.0 /Users/the-gaboardi/miniconda3/envs/py311_spopt/bin/python3.11

self = <spopt.tests.test_region_k_means.TestRegionKMeansHeuristic object at 0x168a95f90>

    @pytest.mark.filterwarnings("ignore:The weights matrix is not fully")
    def test_region_k_means_heuristic_synth_small(self):
        model = RegionKMeansHeuristic(
            self.data_small, self.reg_small, self.w_small, seed=RANDOM_STATE
        )
        model.solve()
    
>       numpy.testing.assert_equal(model.labels_, self.known_labels_small)

spopt/tests/test_region_k_means.py:45: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (<built-in function eq>, array([1, 0, 0, 0, 1, 2, 1, 0, 2]), [1, 1, 1, 0, 1, 2, ...])
kwds = {'err_msg': '', 'header': 'Arrays are not equal', 'strict': False, 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 3 / 9 (33.3%)
E           Max absolute difference: 1
E           Max relative difference: 1.
E            x: array([1, 0, 0, 0, 1, 2, 1, 0, 2])
E            y: array([1, 1, 1, 0, 1, 2, 0, 0, 2])

../miniconda3/envs/py311_spopt/lib/python3.11/contextlib.py:81: AssertionError
```

cc @knaaptime

[...]

PR #353, opened 2023-02-26
Overhaul repo infrastructure

### LINKED ISSUE #325: adopt `pyproject.toml`
We should explore converting to `pyproject.toml` (as in [`spaghetti`](https://github.com/pysal/spaghetti/blob/main/pyproject.toml)), which significantly streamlines repo structure.

jGaboardi:
This PR:
* resolves #325 
* adopts `pyproject.toml` for packaging
* drop `versioneer` in favor of `setuptools_scm` for version control
* adopts `ruff` for linting and lints repo
* cleans up testing ends (see [here](https://github.com/pysal/libpysal/pull/516#pullrequestreview-1299183101))
* removes all redundant files


.ci/311-DEV.yaml
```diff
@@ -3,19 +3,21 @@ channels:
   - conda-forge
 dependencies:
   - python=3.11
-  - geopandas
```
martinfleis: I like to keep the packages installed from main also in here because this ensures that their dependencies get installed from conda. If you keep them only below, shapely and other stuff will come from PyPI.

jGaboardi: Ah ha.... This is a very good point.

jGaboardi: d18503c

jGaboardi:
@martinfleis Any chance for a review on this? 99.99% infrastructure/formatting changes and no changes in code functionality.

.github/workflows/testing.yml
```diff
@@ -48,12 +47,7 @@
 
        - name: install pulp via pip
          shell: bash -l {0}
-         run: pip install pulp 
-       
-       - name: install bleeding edge libpysal (only Ubuntu w/ latest Python)
-         shell: bash -l {0}
-         run: pip install git+https://github.com/pysal/libpysal.git@master
-         if: matrix.os == 'ubuntu-latest' && contains(matrix.environment-file, 'DEV')
+         run: pip install pulp
```
martinfleis, line 50: Why is this here and not in the env yaml files?

martinfleis (commented):
One note in code. 

The CI failure is expected?

Can you make sure that pre-commit.ci is green?

.github/workflows/testing.yml
jGaboardi: Previously there was problem with `pulp` being installed correctly from `conda`, even when done within the `pip` block. It had to done with only some of the available solvers being initialized. This would be a good time to revisit the process. Let me check that out.

jGaboardi:
> The CI failure is expected?

The CI failure is recorded in #346 and only started with https://github.com/pysal/libpysal/pull/511. I haven't been able to track down the reason.

jGaboardi:
> Can you make sure that pre-commit.ci is green?

Oh man. Not sure how this fell through the cracks. The PR had been passing `ruff`+`black` locally for me, not sure how all these got reverted. I'll look into it.

jGaboardi:
`pre-commit.ci` is now back to green.

jGaboardi:
ugh... many more failures after ruffing.

jGaboardi:
@martinfleis All tests back to green (except the 'expected' 3.11-DEV' failure) after my `ruff`ing debacle. Thanks for your patience.

.ci/311-DEV.yaml
```diff
@@ -3,7 +3,10 @@ channels:
   - conda-forge
 dependencies:
   - python=3.11
-  - geopandas>=0.10.0
+  - geopandas
+  - folium
```
martinfleis, line 7: Do we use folium here? In any case, it will come with geopandas from conda.

spopt/region/azp.py
```diff
@@ -436,7 +438,7 @@ def fit_from_dict(
             corresponding area is assigned to at the beginning of the
             algorithm.
             If None, then a random initial clustering will be generated.
-        objective_func : :class:`region.ObjectiveFunction`, default: ObjectiveFunctionPairwise()
```
martinfleis, line 439: Why have you removed this link?

martinfleis (approved):
looks good. thanks!

.ci/311-DEV.yaml
jGaboardi: Ah, forgot to remove that when I added geopandas back into the `conda install` portion of the environment. Will take care of that.

.github/workflows/testing.yml
jGaboardi: NOTE TO SELF -- I still need to look into this.

spopt/region/azp.py
jGaboardi: Can't remember. I will add back and update the others in `azp`

PR #360, opened 2023-03-21
address new `ruff` failures

jGaboardi:
~~***EDIT*** On hold until more insight from https://github.com/pysal/momepy/pull/476.~~

This PR:
* follows up #359 with fixes for the fresh `ruff` failures.
* supersedes #361

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 files: 'spopt\/'
 repos:
   - repo: https://github.com/psf/black
-    rev: 23.1.0
+    rev: 23.3.0
     hooks:
       - id: black
         language_version: python3
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.0.255"
+    rev: "v0.0.261"
     hooks:
       - id: ruff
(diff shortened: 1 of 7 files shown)


martinfleis (commented):
I'll leave the decision on how precise you want stack levels to be to you but wanted to flag that `stacklevel=1` is always wrong one (not informative), so you may want to revisit it if you thought it has proper levels. We should generally try to show the level of a user call to a function that caused the warning. E.g.

```py
from warnings import warn

def fn():
    warn("my warn", stacklevel=2)

fn()
```

which shows 
```
.../file.py:1: UserWarning: my warn
  fn()
```

That tells you that the warning is caused by a call to `fn()`. While `stacklevel=1` shows

```
.../file.py:1: UserWarning: my warn
  warn("my warn", stacklevel=1)
```

which is not useful.

jGaboardi:
Ah. I'll give this some more thought.

jGaboardi:
@martinfleis OK, I think I understand this now and that proper stack levels are being applied.

martinfleis (approved):
Looking good!

Issue #364, opened 2023-04-20
Consider using pulp.Binary rather than pulp.Integer as the type for cli_assgn_vars?


ljwolf:
Hi,

On some data I cannot share (sorry!) the P-Median problem solved in CPLEX on Windows is returning assignments that are *nearly* zero or one, but are not quite: 

![index_confusions](https://user-images.githubusercontent.com/2250995/233401787-b825b48d-cf8a-4043-be03-0c1c72bdb710.PNG)

This results in `model.cli2fac` having multiple assignments for some clients, when the P-Median model is supposed to be singly-assigned. We do not yet support partial coverage, and our constraints should prevent this behaviour. To be clear, I think this is CPLEX's problem, not ours. 

At the same time—I know CPLEX is trying its best, but I wonder whether we can help them out by using `pulp.Binary` rather than `pulp.Integer` for binary decision variables? I think this could be done by replacing the `pulp.Integer` types with `pulp.Binary` throughout `FacilityModelBuilder`, and dropping the upper and lower bounds.

ljwolf:
Yes, can try. 

I think, too, we may want to implement a more robust way of constructing `fac2cli` and `cli2fac` in light of these kinds of precision issues... in model instances we know are singly-assigned, maybe we should round `cli_assgn_vars` to an integer when using its values? 

In light of this, I managed to replicate this behaviour with Gurobi in a P-Median on different data by setting the optimality gap to some low number to force early termination at a bad solution. I wonder whether this is more common than we think?

jGaboardi:
Maybe as an order of things to try/implement:
1. @ljwolf's rounding solution
2. See about swapping out integer for binary 
3. manage the logic internally to distinguish between scenarios needing relaxation (capped LSCP) and those that do not (standard PMP instances).

jGaboardi:
Resolved via #366

PR #366, opened 2023-04-25
default to binary assignment for clients

jGaboardi:
This PR (hopefully) provides a solution to #364. Simply swapping out `pulp.LpInteger` for `pulp.LpBinary` as the default in [`add_client_assign_variable`](https://github.com/pysal/spopt/blob/main/spopt/locate/base.py#L241) seems to do the trick and tests pass locally. @ljwolf Can you try this out on your data to see if it resolves the issue?

### DIFF
--- a/spopt/locate/base.py
+++ b/spopt/locate/base.py
@@ -238,7 +238,7 @@ class FacilityModelBuilder:
         var_name: str,
         low_bound=0,
         up_bound=1,
-        lp_category=pulp.LpInteger,
+        lp_category=pulp.LpBinary,
     ) -> None:
         """Client assignment integer decision variables (used for allocation).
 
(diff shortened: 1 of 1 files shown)


gegen07 (approved):

ljwolf:
Yes! Looks fine! 

I don't know if it will solve the problem generally, as I'm away from my windows laptop for a while that reproduced the problem.

jGaboardi:
> Yes! Looks fine!
> 
> I don't know if it will solve the problem generally, as I'm away from my windows laptop for a while that reproduced the problem.

OK, shall we hold off merging until we can verify?

jGaboardi:
CI failures are not related to this PR – see https://github.com/pysal/spopt/issues/386

jGaboardi:
@ljwolf I think there is no harm in merging this, even if we can't verify it fixes your specific use case. Thoughts?

ljwolf:
Indeed! Fine to merge this!

PR #374, opened 2023-05-26
capacitated p-median with predefined facilities

rongboxu:


spopt/locate/base.py
```diff
@@ -449,37 +453,83 @@ def add_facility_constraint(obj: T_FacModel, p_facilities: int) -> None:
...
 
         Returns
         -------
 
         None
 
         """
+        if predefined_fac.ndim == 2:
+            n, k = predefined_fac.shape
+            assert (
```
ljwolf, line 484: Change this to raise a `ValueError()` rather than `assert`.

spopt/locate/base.py
```diff
@@ -449,37 +453,83 @@ def add_facility_constraint(obj: T_FacModel, p_facilities: int) -> None:
...
+                k == 1
+            ), "predefined facilties array must only be of shape (n_supply, 1) or (n_supply,)"
```
ljwolf, line 486: try something like: 
```python
if k!=1:
    ValueError(
    "predefined facilities array must only be of shape"
    " (n_supply, 1) or (n_supply,)"
    )
```

spopt/locate/base.py
```diff
@@ -449,37 +453,86 @@ def add_facility_constraint(obj: T_FacModel, p_facilities: int) -> None:
...
+        facility_capacity : numpy.array (default None)
+            The capacity of each facility.
 
         Returns
         -------
 
         None
 
         """
+        if predefined_fac.ndim == 2:
```
jGaboardi, line 482: Nice job on the exception handling you have added in here (and elsewhere).

spopt/locate/p_median.py
```diff
@@ -246,10 +253,50 @@ def from_cost_matrix(
...
+                )
+                if sum_predefined_fac_cap <= weights.sum():
+                    FacilityModelBuilder.add_predefined_facility_constraint(
+                        p_median,
+                        predefined_facilities_arr,
+                        weights,
+                        facility_capacities,
+                    )
+                else:
+                    raise SpecificationError(
```
jGaboardi, line 268: Nice job on the test checking that the `SpecificationError` is being raised appropriately. In the future we will want to test that all the new exceptions are raised when appropriate, but this is a great start!

spopt/tests/data/example_subject_schools.csv
```diff
@@ -0,0 +1,75 @@
+,SE2 PP: Code,SE2 PP: PC,PL: Subject,Count,priority
```
jGaboardi, line 1: I am wondering if it is OK to be using this IOE data here (for privacy, etc.). How much of this example school data is synthetic? In general, I think it is great to have real-world tests. However, the output in `example_subject_student_school_journeys.csv` is rather large to be storing in the repo.

@ljwolf & @gegen07 Any thoughts?

spopt/tests/test_c_p_median.py
```diff
@@ -0,0 +1,227 @@
...
+import pulp
+import spaghetti
+
+from spopt.locate import PMedian
+from spopt.locate.util import simulated_geo_points
+import pytest
+import os
+
+
+class TestSyntheticLocate:
```
jGaboardi, line 19: Excellent, small synthetic test.

spopt/tests/data/example_subject_schools.csv
rongboxu: It should be all synthetic. Nick also shows the data in a public repository for IOE's case. I can ask him for confirmation later!

gegen07: Maybe we can think about a repo to retrive the datasets that are used in our tests. I think `libpysal.examples` is deprecated, right? We can open a PR to include the datasets to [geodatasets](https://github.com/geopandas/geodatasets) repo which is under geopandas org. What do you think?

jGaboardi: Hmmm. Yeah, it's listed in [items to deprecate](https://github.com/pysal/libpysal/issues/484). But there's also [continuing activity in that regard](https://github.com/pysal/pysal/issues/1276). Perhaps we can get @sjsrey's input here?

jGaboardi: @ljwolf What do we think about keeping the testing data as is for this PR and opening a ticket for potential updating in the future? The 3 `.csv` files are a combined ≈187 KB, so def not back-breakers.

jGaboardi:
We should also open a tickets:
* to add an example notebook demonstrating this new functionality
* adding any references to formulations @rongboxu used (if not already in the [docs](https://pysal.org/spopt/references.html))
* anything else?

To be clear, these things are not critical to be part of this PR.

rongboxu:
> We should also open tickets:
> 
> * to add an example notebook demonstrating this new functionality
> * adding any references to formulations @rongboxu used (if not already in the [docs](https://pysal.org/spopt/references.html))
> * anything else?
> 
> To be clear, these things are not critical to be part of this PR.

Thank you for opening these!

* About 1, I think I can do it by taking the existing notebook as an example!
* About 2, yes, I need to add one preference, which is the book, business site selection, location analysis and GIS.

jGaboardi:
> * About 2, yes, I need to add one preference, which is the book, business site selection, location analysis and GIS.

I thought we had already had that one in our [*References*](https://pysal.org/spopt/references.html), but it looks we had missed it. So nice catch. A `bibtex` entry will need to be added to [`references.bib`](https://github.com/pysal/spopt/blob/main/docs/_static/references.bib)

Issue #348, opened 2023-02-06
more detailed instructions for `pulp`/`cbc` install


jGaboardi:
Following up from #347, we should add more detailed instructions for `pulp`/`cbc` installation in:

* `README.md`
* `docs/installation.rst`

cc @qszhao

gegen07:
From the discussion on [PR](https://github.com/pysal/libpysal/pull/666#discussion_r1426901833). Is it acceptable to loop over the solver available list with a priority list like (get GUROBI if it is available, then COIN, then...). I think PULP has a default installation, but unfortunately, we cannot set the default solver to be installed.

This solution has some warnings and errors to be raised. If the desired solver is not installed, we can raise a warning. If no solver is installed, we raise an error. 

This sounds good? @jGaboardi @ljwolf 

Happy to open a PR, if it is ok.

qszhao:
I think this is a good idea since pulp/cbc can be the default installation on different platforms.

jGaboardi:
@gegen07 Yeah, let's try it. But I think we should stick with open source options as a top defaults.

PR #377, opened 2023-05-27
shapely>=2 & other minimum version deps

### LINKED ISSUE #373: Failing CI: geopandas + spaghetti
The release of [`geopandas==0.13.0`](https://github.com/pysal/spopt/actions/runs/5051951698/jobs/9064313864#step:5:28) is resulting in 58 failures with the following [message](https://github.com/pysal/spopt/actions/runs/5051951698/jobs/9064313864#step:6:4439):

```
AttributeError: You are calling a geospatial method on the GeoDataFrame, but the active geometry column to use has not been set. 
There are no existing columns with geometry data type. You can add a geometry column as the active geometry column with df.set_geometry.
```

These are ultimately tied back to [`spaghetti`](https://github.com/pysal/spaghetti), whose `main` is passing, but needs a fresh release. Prior to that release, I would like to see https://github.com/pysal/spaghetti/pull/721 and https://github.com/pysal/spaghetti/pull/724 resolved.

xref #371

### LINKED ISSUE #375: stipulate minimum `shapely` version
In `spopt` we are actually never stipulating a minimum supported version of `shapely`[^1][^2]. This is now causing a [CI failure](https://github.com/pysal/spopt/actions/runs/5074939278/jobs/9138264361#step:6:5683) by resolving to [`shapely==1.8.0`](https://github.com/pysal/spopt/actions/runs/5074939278/jobs/9138264361#step:5:32), which is no longer [supported by `spaghetti`](https://github.com/pysal/spaghetti/blob/c25f662c26ce9a909101ca879c681b3a91e0904b/pyproject.toml#L65) leading to:

```
 AttributeError: module 'shapely' has no attribute 'points'
```

I vote that we bump minimum supported versions as follows:
* `geopandas>=0.12`
* `shapely>=2`

Thoughts? 

cc @sjsrey @ljwolf @gegen07 @martinfleis @knaaptime 

[^1]: [`pyproject.toml`](https://github.com/pysal/spopt/blob/21efef3da7253598047b0185c8944b65b56f8063/pyproject.toml#L32)
[^2]: [`/.ci/38-MIN.yaml`](https://github.com/pysal/spopt/blob/21efef3da7253598047b0185c8944b65b56f8063/.ci/38-MIN.yaml#L14)

### LINKED ISSUE #376: update micromamba setup in CI
xref https://github.com/pysal/spaghetti/issues/727

jGaboardi:
This PR:
* resolves #375 
* resolves #376 
* resolves #373


jGaboardi:
Looks like even `python==3.8` is causing problems. Perhaps time to dump that and support only `python>=3.9`.

Thoughts? 

cc @sjsrey @martinfleis @gegen07 @knaaptime @ljwolf

sjsrey:
In this specific case that makes sense. More broadly we might think of adopting a policy of supporting say the last three stable versions of Python?

If we did that, we'd have a decision rule as to when to drop previous versions on a case by case basis.

jGaboardi:
No idea why I am not able to get this 3.9 minimum CI environment to solve...

martinfleis:
We should follow https://scientific-python.org/specs/spec-0000/ in regards to versions. Including Python. We can deviate when it makes sense (shapely 2).

jGaboardi:
I finally realized the problem was specifically with the `geopandas==0.12` pin, which was not necessary. Keeping the pin at `geopandas==0.10` allows for Python 3.8 environment resolution.

.ci/38-MIN.yaml
```diff
@@ -1,17 +1,17 @@
-name: test
+name: py38_spopt
```
martinfleis, line 1: I guess you changed this for testing locally? Can you revert?

.ci/38-MIN.yaml
```diff
@@ -1,17 +1,17 @@
...
   - libpysal
   - matplotlib
-  - networkx
+  - networkx=2.6
   - numpy=1.21
-  - pandas=1.0.5
-  - scikit-learn=0.22
-  - scipy=1.3.2
-  - shapely
+  - pandas>=0.23.0,!=1.5.0,<2
```
martinfleis, line 11: I would potentially just pin a specific version here.

.github/workflows/testing.yml
```diff
@@ -33,43 +31,39 @@
...
          with:
            environment-file: ${{ matrix.environment-file }}
            micromamba-version: 'latest'
-           channel-priority: 'flexible'
         
        ##############################################################################
        #- name: install pulp via pip
-       #  shell: bash -l {0}
-       #  run: pip install pulp
+       #  run: |
```
martinfleis, line 51: What is this?

.ci/38-MIN.yaml
jGaboardi: So, I modeled that after the ["pin" we have in `momepy`](https://github.com/pysal/momepy/blob/fbf2c2fbfc95a133c9cebcb35eb59e169a1bcf5c/ci/envs/38-minimal.yaml#L13). Perhaps we should update that, as well? Here' let's try a [`pandas==1.3`]() to conform with the [SPEC 0 — MINIMUM SUPPORTED VERSIONS](https://scientific-python.org/specs/spec-0000/) you posted. I will update the `pyproject.toml` accordingly.

.github/workflows/testing.yml
jGaboardi: This something that I want to keep in for now in case we need to uncomment in the future. Basically, `pulp`'s solver installation is a bit varied by OS and installation method. Sometimes previously we need to run `pulptest` to get the proper solvers installed and available following actual `pulp` installation. xref #348

martinfleis (approved):

Issue #386, opened 2023-06-19
`TestAZP.test_azp_basic_from_w` CI failure


jGaboardi:
[`TestAZP.test_azp_basic_from_w`](https://github.com/pysal/spopt/blob/aa6e37933b6e46373f21888e9ff380c66d7e8145/spopt/tests/test_azp.py#L31) CI [failure](https://github.com/pysal/spopt/actions/runs/5306722283/jobs/9604584528#step:6:504).

#### diagnosis & potential causes
* Seems to fail intermittently across versions and OSs (fails locally on MacOS) ***with the same resultant array***
* Could be an issue with the random state?
* Could be a change in `numpy` or `scipy`?

#### major versions
* `geopandas==0.13.2`
* `networkx== 3.1`
* `numpy==1.25.0`
* `pandas==2.0.2`
* `scipy==1.10.1`
* `shapely==2.0.1`

```
=========================================================================== FAILURES ============================================================================
_________________________________________________________________ TestAZP.test_azp_basic_from_w _________________________________________________________________
[gw0] darwin -- Python 3.11.4 /Users/user/mambaforge/envs/py311_spopt/bin/python3.11

self = <spopt.tests.test_azp.TestAZP object at 0x162099350>

    def test_azp_basic_from_w(self):
        w = libpysal.weights.Queen.from_dataframe(self.mexico)
        attrs_name = [f"PCGDP{year}" for year in range(1950, 2010, 10)]
        args = (self.mexico, w, attrs_name)
        kwargs = {"n_clusters": 3, "random_state": RANDOM_STATE}
        model = AZP(*args, **kwargs)
        model.solve()
    
>       numpy.testing.assert_array_equal(model.labels_, self.basic_from_w_labels)

spopt/tests/test_azp.py:39: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (<built-in function eq>, array([0., 0., 0., 0., 1., 1., 1., 2., 0., 1., 1., 0., 0., 2., 2., 2., 2.,
[...]

PR #403, opened 2023-10-14
remove `zip(strict=True)` in k-nearest p-median

### LINKED ISSUE #402: k-nearest pmedian failures due to `zip(strict=True)`
[2023-10-13 CI – py38](https://github.com/pysal/spopt/actions/runs/6514516150/job/17695823053#step:6:630)

k-nearest pmedian failures due to `zip(strict=True)`

```
FAILED spopt/tests/test_knearest_p_median.py::TestKNearestPMedian::test_knearest_p_median_from_geodataframe - TypeError: zip() takes no keyword arguments
FAILED spopt/tests/test_knearest_p_median.py::TestKNearestPMedian::test_knearest_p_median_from_geodataframe_no_results - TypeError: zip() takes no keyword arguments
FAILED spopt/tests/test_knearest_p_median.py::TestKNearestPMedian::test_solve - TypeError: zip() takes no keyword arguments
```

This seems to have flown under the radar due to #386. I will remove the `strict=True` keyword for now. We can re-add once `python=3.{8,9}` are dropped and python=3.12 is added , which will be soon. I anticipate this for a `v0.6.1` release.

jGaboardi:
This PR resolves #402 and runs `pmedian.ipynb` for output.


jGaboardi:
Can confirm that all tests are passing/failing as they should (considering #386).

spopt/locate/p_median.py
```diff
@@ -806,8 +806,10 @@ def _from_sparse_matrix(self) -> None:
 
         # Client assignment integer decision variables
         row_indices, col_indices, values = find(self.aij)
+
+        # TODO: add `strict=True` back in to `zip` – gh-spopt#402
```
martinfleis, line 810: I'd remove this. We're zipping an output of scipy.sparse.find. That is never of different length so using strict here has no reason.

jGaboardi: Good point. I'll remove the note and add a `# noqa`

martinfleis: Why `noqa`?

jGaboardi: > Why `noqa`?

[`B905`](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) is raised without it.

jGaboardi: @martinfleis Is there something I am missing that makes this a bad idea?

martinfleis: > [B905](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) is raised without it.

Is it? I don't see it locally and it should not, given we specify `target-version = "py38"`. Ruff should not suggest to include a keyword that does not exist in 3.8.

martinfleis: It doesn't matter in practice, it is just strange that you see this. I don't mind to have it there, I just think that right now it should not raise and once we update min Python, Ruff will tell us to add a keyword here. But not now.

jGaboardi: Fair enough. I am being proactive with it. No need to do so, I suppose. I'll remove and merge.

martinfleis (approved):
Thanks!

PR #406, opened 2023-10-17
Fix CI failure due to AZP multiple valid MST

gegen07:
## Summary

- Read the issue comment [here](https://github.com/pysal/spopt/issues/386#issuecomment-1766514328). TLDR; the MST of `scipy.sparse` graphs generates some fluctuations. Then the proposed fix is presented in `util.py`. Working with an undirected graph give the same results over the python versions.

- MST of an unweighted graph is all trees (a path with a single component). Then it would give more than one valid MST using the `scipy` algorithm. Thus, I computed the distance between centroid polygons based on Queen Weight contiguity to generate a unique valid MST based on Mexico shapefile in `test_azp.py`.

- I changed the expected labels in the test to consider all work done in the PR.


spopt/tests/test_azp.py
```diff
@@ -30,12 +30,26 @@ def setup_method(self):
 
     def test_azp_basic_from_w(self):
         w = libpysal.weights.Queen.from_dataframe(self.mexico)
+
+        weights = {}
+        for k, v in w.neighbors.items():
+            arr = []
+            x = MEXICO.iloc[k].geometry.centroid
```
jGaboardi, line 37:
```suggestion
            x = self.mexico.iloc[k].geometry.centroid
```

spopt/tests/test_azp.py
```diff
@@ -30,12 +30,26 @@ def setup_method(self):
...
+            for n in v:
+                y = MEXICO.iloc[n].geometry.centroid
```
jGaboardi, line 39:
```suggestion
                y = self.mexico.iloc[n].geometry.centroid
```

jGaboardi:
@gegen07 This is great work in finding the exact cause of the discrepency here. While the solution here is probably not the most efficient, it certainly does get the job done and makes CI green again (which is a huge accomplishment).

Here's a couple of points to start off the discussion:
* > I computed the distance between centroid polygons based on Queen Weight contiguity to generate a unique valid MST
  * So will this distance method need to be added into `AZP` itself to ensure user results are reproducible?
* I am curious how the new [`Graph` class](https://github.com/pysal/libpysal/blob/main/libpysal/graph/base.py#L35) might fit in here.

@knaaptime @martinfleis What are yalls thoughts on this?

jGaboardi:
I'm also wondering if functions like [`generate_initial_sol()`](https://github.com/pysal/spopt/blob/76553409db2e358c1772f0d70a17c0ef4ea6ae8c/spopt/region/util.py#L646) might be better place within [`azp_util.py`](https://github.com/pysal/spopt/blob/76553409db2e358c1772f0d70a17c0ef4ea6ae8c/spopt/region/azp_util.py) since it's only used for AZP stuff.

jGaboardi:
New failures due to https://github.com/pysal/libpysal/issues/605, https://github.com/pysal/esda/issues/271. Fix is on the way.

spopt/region/util.py
```diff
@@ -747,6 +743,14 @@ def _randomly_divide_connected_graph(adj, n_regions):
             f"equal to the number of nodes which is {n_areas}."
         )
     mst = csg.minimum_spanning_tree(adj)
+    mst_copy = mst.copy()
```
ljwolf, line 746: I think you can use a classic trick here: 

`symmetric_mst = (mst + mst.T) > 0`

jGaboardi: @ljwolf How does this look?

gegen07: This is so smooth!

jGaboardi: @gegen07 I know, right? Always some tricks to learn that I have never seen!

jGaboardi (approved):
I think this is ready to go in my opinion, but we'll wait for @ljwolf' s approval to see if there are any more easy improvements.

PR #408, opened 2023-10-24
Update 311-DEV.yaml

### LINKED ISSUE #407: add `esda` bleeding edge to `DEV` testing
@martinfleis Do you see any drawback to adding the bleeding edge `esda` to [`ci/311-DEV.yaml`](https://github.com/pysal/spopt/blob/main/ci/311-DEV.yaml)?

jGaboardi:
This PR resolves #407

### DIFF
--- a/ci/311-DEV.yaml
+++ b/ci/311-DEV.yaml
@@ -25,8 +25,9 @@ dependencies:
   - pip
   - pip:
     - pulp
+    - git+https://github.com/pysal/esda.git
     - git+https://github.com/geopandas/geopandas.git
-    - git+https://github.com/networkx/networkx.git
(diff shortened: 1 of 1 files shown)


martinfleis:
But spopt does not seem to depend on esda... Am I missing anything? Is that a dependency via another package?

jGaboardi:
> But spopt does not seem to depend on esda... Am I missing anything? Is that a dependency via another package?

Yes, `spaghetti`.

martinfleis:
Not sure how deep we need to go but it won't hurt for sure.

jGaboardi:
Yeah, if the failure wasn't in the PySAL ecosystem then I wouldn't be as for this.

Issue #410, opened 2023-12-02
client demand and facility capacity is transposed in `LSCP.from_geodataframe()`


ljwolf:
In the `from_cost_method()` function definition, demand weight comes first: 

https://github.com/pysal/spopt/blob/main/spopt/locate/coverage.py#L126

However, `from_geodataframe()` passes the facility capacities first: 

https://github.com/pysal/spopt/blob/main/spopt/locate/coverage.py#L455

This makes it hard to use the capacities, since any situation where capacity is bigger than the demand will get rejected erroneously as if capacity were *smaller* than the demand. 

A fix is both easy and immediate. I can make this change, and would really appreciate a bug fix release this week before teaching the module in Tartu?

jGaboardi:
Resolved via #411 – will close issue following release.

PR #411, opened 2023-12-02
solve b410

ljwolf:
Bug #410 means that no capacitated LSCP will ever be solvable through `from_geodataframe()`. The check requires that demand weight is less than capacity, but transposes the capacity and demand arrays. So, all cases where capacity is greater than demand have an error raised saying that there is not enough capacity, while all cases where capacity is *less than* demand proceed to solve and return infeasible. 

This fixes the transposition.


jGaboardi:
Looks like no tests are affected by this. Probably means we need better tests for it.

jGaboardi (approved):

PR #418, opened 2023-12-04
stricter linting – follow up #413

jGaboardi:
* Stricter linting – follow up #413
* This touches more lines of code than I thought it would... As usual, I can split if desired.

### DIFF
--- a/spopt/locate/base.py
+++ b/spopt/locate/base.py
@@ -1,3 +1,5 @@
+# ruff: noqa: B009, B010
+
 from abc import abstractmethod
 from typing import TypeVar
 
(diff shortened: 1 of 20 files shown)


spopt/locate/base.py
```diff
@@ -1,3 +1,5 @@
+# ruff: noqa: B009, B010
```
gegen07, line 1: The ruff rules are assigned how? Just to know the reasons.

jGaboardi: This is a good question. The `ruff` config options can be found [here](https://docs.astral.sh/ruff/configuration/) and the rules can be found [here](https://docs.astral.sh/ruff/rules/). I am not sure there is a definitive guide, but my experience with these linters makes me want to be as granular as possible. This means:
* limiting super-global rules found in `pyproject.toml` to a bare minimum
* preferring in-line `noqa` to the file-level `noqa` rules if there are only a few (and this is ambiguous) instances
* using file-level `noqa` rules especially on legacy code.

I think for writing new code we should try to limit ourselves to in-line `noqa` where possible unless there is a good reason not to.

That's my 2-cents on the topic. Happy to hear yours, of course, @gegen07 & @martinfleis

spopt/region/maxp.py
```diff
@@ -154,11 +154,11 @@ def maxp(
 
 def construction_phase(
```
martinfleis, line 155: Make this and other internal functions private (e.g. `_construction_phase`)

martinfleis (approved):

PR #420, opened 2023-12-09
Formulation of KN-PMP omitted opening constraints 

### LINKED ISSUE #419: no opening constraints in $k$-nearest $p$-median
After noticing some strange results in the $k$-nearest $p$-median implementation, I have realized that we had completely [omitted the generation of opening constraints in the model](https://github.com/pysal/spopt/blob/bf7134a33794e9a02377e95563faaa48d74a486d/spopt/locate/p_median.py#L847). While the model will still solve (and potentially optimally) in certain scenarios, the results are erroneous:
* some clients may not be allocated to a facility
* the $p$ constraint not respected, since clients can be allocated to any facility

I have verified with diagnostics and plotting.

jGaboardi:
This PR resolves #419

### DIFF
--- a/spopt/locate/base.py
+++ b/spopt/locate/base.py
@@ -749,8 +749,11 @@ class FacilityModelBuilder:
             fac_vars = getattr(obj, "fac_vars")
             model = getattr(obj, "problem")
 
+            from_dict = isinstance(cli_assgn_vars, dict)
             for i in range_client:
                 for j in range_facility:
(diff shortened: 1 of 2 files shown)


ljwolf:
This looks good to me! It appears that without the opening constraints, clients weren't getting totally assigned.

jGaboardi:
> This looks good to me! It appears that without the opening constraints, clients weren't getting totally assigned.

Yep, and being assigned to closed facilities.

Issue #414, opened 2023-12-03
`v0.6.0` release checklist


jGaboardi:
* [x] #364
* [x] #366 
* [x] #401 
* [x] #406 
* [x] #410
* [x] #411
* [x] #412
* [x] #413
* [x] #418
* [x] #416 
* [x] #417
* [x] #419 
* [x] #420 
* [x] #421 
* [x] #422 
* [x] #423
* [x] #424
* [x] #425
* [x] #426
* [x] #427 
* [x] #429 
* [x] DOCS
   * [x] add KNPMP stuff to API & Tutorials
   * [x] rerun all notebooks
   * [x] make sure docs build locally

jGaboardi:
Going to keep this open until all all issues are resolved

PR #424, opened 2023-12-10
parameters & docstring clean up in `KNearestPMedian`

### LINKED ISSUE #423: clean up `KNearestPMedian` parameters & docstring
The parameters in [`KNearestPMedian`](https://github.com/pysal/spopt/blob/bf7134a33794e9a02377e95563faaa48d74a486d/spopt/locate/p_median.py#L546) diverge a bit from the other facility model classes. This should be a simple fix + cleaned up docstrings. And since `KNearestPMedian` has not yet been part of a release, there will be no need for a deprecation period.

jGaboardi:
This PR resolves #423.

xref #414

### DIFF
--- a/spopt/locate/p_median.py
+++ b/spopt/locate/p_median.py
@@ -593,14 +593,14 @@ class KNearestPMedian(PMedian):
         An array of coordinates of facilities.
     weights : np.array
         An array of weights representing the service loads of the clients.
+    k_array : np.array
+        An array of k values representing the number of nearest facilities
+        for each client.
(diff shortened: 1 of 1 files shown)


spopt/locate/p_median.py
```diff
@@ -641,7 +642,7 @@ def __init__(
         distance_metric: str = "euclidean",
         name="k-nearest-p-median",
```
jGaboardi, line 643: This line needs to be removed
```suggestion
```

PR #431, opened 2023-12-10
K-nearest p-median demo notebook

jGaboardi:
This PR is the final one before cutting the `v0.6.0` release. Although it's a huge diff, nearly all is either rerun notebooks or the new notebook.

xref #414


ljwolf:
This looks ok to me, but I haven’t checked the specific notebooks. I’m keen to clear the decks for a release, so I’d like to merge this!

PR #449, opened 2024-04-27
testing code cleanup

### LINKED ISSUE #448: better raises/warns control + DRY out testing
* Need to improve betterwarns control in testing to cut back on thrown warnings in CI
* Need to iDRY out `pytest.{raises,warns}` and move to `conftest.py`

xref: #437

### LINKED ISSUE #450: "Generate and publish the report" not working - remove
["Generate and publish the report"](https://github.com/pysal/spopt/blob/785b396d8459491fee6e5a826b02b513fabf89f3/.github/workflows/testing.yml#L93) not working - remove

### LINKED ISSUE #451: `codecov` acting up a bit – update action
* `codecov` seems to be [acting up](https://github.com/pysal/spopt/actions/runs/8862908203/job/24336386929?pr=449#step:7:34) a bit.
* try to update the action [like `momepy`](https://github.com/pysal/momepy/blob/8e93b6c610313d7b07eb564564c0caea5600c0df/.github/workflows/tests.yaml#L60)

### LINKED ISSUE #452: `numpy` DeprecationWarning within `pyproj`
A `numpy` DeprecationWarning is being thrown from within `pyproj` (see [here](https://github.com/pysal/spopt/actions/runs/8864071815/job/24338864716?pr=449#step:6:386). Is this something we should ignore or do something about on our side?

xref:
* https://github.com/pyproj4/pyproj/issues/1307
* #449 

```
spopt/tests/test_lscp.py: 6 warnings
spopt/tests/test_lscpb.py: 6 warnings
spopt/tests/test_mclp.py: 8 warnings
spopt/tests/test_p_center.py: 6 warnings
spopt/tests/test_p_median.py: 8 warnings
  /home/runner/micromamba/envs/test/lib/python3.12/site-packages/pyproj/transformer.py:820: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    return self._transformer._transform_point(
```

jGaboardi:
This PR:
* resolves #448 
* resolves #450 
* resolves #451 
* resolves #452
* eliminated CI warnings in all by `oldest` and `dev`

### DIFF
--- a/spopt/tests/test_clscp-so.py
+++ b/spopt/tests/test_clscp-so.py
@@ -1,5 +1,4 @@
 # ruff: noqa: N999
-
 import numpy
 import pulp
 import pytest
--- a/spopt/tests/test_clscp-so.py
+++ b/spopt/tests/test_clscp-so.py
@@ -94,7 +94,7 @@ class TestSyntheticLocate:
                 demand_quantity_arr=demand_quantity,
             )
 
-    def test_clscpso_infease_error(self):
+    def test_clscpso_infease_error(self, loc_raises_infeasible):
         service_radius = 1
         facility_capacity = numpy.array([5, 15])
         demand_quantity = numpy.arange(1, 6)
@@ -104,5 +104,5 @@ class TestSyntheticLocate:
             facility_capacity_arr=facility_capacity,
             demand_quantity_arr=demand_quantity,
         )
-        with pytest.raises(RuntimeError, match="Model is not solved:"):
+        with loc_raises_infeasible:
             clscpso.solve(pulp.PULP_CBC_CMD(msg=False))
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -76,7 +76,7 @@
            pytest spopt \
            -v \
            -r a \
-           -n auto \
+           -n logical \
            --color yes \
            --cov spopt \
            --cov-append \
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -81,21 +81,7 @@
            --cov spopt \
            --cov-append \
            --cov-report term-missing \
-           --cov-report xml
+           --cov-report xml .
        
        - name: codecov
-         uses: codecov/codecov-action@v3
-         with:
-           token: ${{ secrets.CODECOV_TOKEN }}
-           file: ./coverage.xml
-           name: spopt-codecov
-
-       - name: Generate and publish the report
-         if: |
-           failure()
-           && steps.status.outcome == 'failure'
-           && github.event_name == 'schedule'
-           && github.repository_owner == 'pysal'
-         uses: xarray-contrib/issue-from-pytest-log@v1
-         with:
-           log-path: pytest-log.jsonl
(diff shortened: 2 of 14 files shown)


spopt/tests/test_clscp-so.py
gegen07, line 1: There is a similar infeasibility warning in line 86. Should we move all expected warnings to `conftest.py`?

gegen07 (commented):
Excellent work @jGaboardi! The PR smoothens the code a bit and adds more reusability.

- I noted some warnings were not moved to `conftest.py`. Is there a specific reason for deciding which test should use the `conftest.py` warning?
- About the numpy `deprecationWarning`, I think we should filter it, as it seems, it will disappear in future releases of `pyproj`.

jGaboardi:
> I noted some warnings were not moved to `conftest.py`. Is there a specific reason for deciding which test should use the `conftest.py` warning?

* Yeah, my general rule was that if identical `warns`/`raises` happen across testing file then put it in `conftest.py`. There is one `raises` that happens twice within `test_knearest` so I just kept that one in the file itself.
* After looking through again, I have missed a few cases. Good catch on that! I'll push up some changes.

spopt/tests/test_clscp-so.py
jGaboardi: > There is a similar infeasibility warning in line 86. Should we move all expected warnings to `conftest.py`?

It's similar, but not the same message. I think keeping where it is now would be best. However, this brings a larger discussion of it we should review and generalize the error messages. If we think that should be the case let's open a fresh issue for that.

jGaboardi:
> About the `numpy` deprecationWarning, I think we should filter it, as it seems, it will disappear in future releases of pyproj.

OK, I will see about adding that to this PR.

jGaboardi:
@gegen07 Seems I missed quite few 😅.

Also, I am having trouble pinning down where the `DeprecationWarning` is being thrown. It happens when I run the tests through `pytest` like usual, but then nothing is thrown after I broke out the code line by line 🤔

jGaboardi:
OK! I think I've found to offending line. Will push up the fix soon.

spopt/tests/test_clscp-so.py
gegen07: I agree with opening a fresh issue. I think we can keep these warnings without generalizing them first. To avoid over-engineering, we can do this transition step-by-step. Firstly, move the warnings in `conftest`. Then, if we see `conftest` has a lot of repeated code, we should generalize it.

gegen07:
> Also, I am having trouble pinning down where the `DeprecationWarning` is being thrown. It happens when I run the tests through `pytest` like usual, but then nothing is thrown after I broke out the code line by line 🤔

I skim the code but did not found what part of the code was throwing the `DeprecationWarning`. Maybe we can just ignore? Does it impact on the report of the test suite?

jGaboardi:
@gegen07 OK. I think this is ready for a proper review.

gegen07 (approved):

Issue #464, opened 2024-11-13
TSP suite? 


ljwolf:
I now have a collection of vehicle routing problem tools built on top of [`pyvrp`](https://github.com/PyVRP/PyVRP) that might be useful for `spopt`. 

The tooling mainly involves data ingestion and conversion to formats/structures that `pyvrp` can natively work with, and then tools to extract geographical output from `pyvrp` solutions. 

So, things like calculating ETAs in timestamps, conversion of input lon/lats to integer positions and back, creating route geometries, etc. 

Is this of interest? I can maintain the public version, but would welcome feedback on API decisions. I think it is substantially different from `locate` as to warrant a new top level `spopt.route` module. 

An example specification might be: 

```python

problem = Route(
    depot_location=(9.2413, 52.2921),
    depot_close=pandas.to_datetime("2025-01-02 20:00:00"),
    depot_name='My Depot',
).add_clients(
    data.geometry,
    data.demand_volume,
    data.time_windows,
    data.source.index,
    data.service_times
).add_trucks_from_frame(
   truck_df
).solve()
```

jGaboardi:
Yes! Something like this was in the initial stages of planning/ideas, but has never come to fruition due to time constraints!

* https://github.com/pysal/spopt/issues/130

jGaboardi:
xref: https://github.com/pysal/spopt/wiki/Architecture#module-structure

jGaboardi:
xref: https://hackmd.io/n7-axu5ITXqjn7mNQNyJlw

jGaboardi:
Some of our initial ideas (from 5 years ago now) label this kind of thing as `route` and some label it as `transport`. I suppose we'll need to determine the most accurate module name.

gegen07:
This is awesome! 

I have another point to think about. We currently use pulp in the `locate` module, and if we introduce `pyvrp` for `route/transport`, that would add another core dependency. This could pose any challenges? How might we manage the environment requirements more smoothly?

Perhaps this would be a good time to meet and discuss our future directions.

jGaboardi:
@ljwolf This is very exciting. Can your stuff built on top of `pyvrp` take network distance/travel time as impedance or is bound to euclidean/haversine distance?

ljwolf:
> can your stuff built on top of `pyvrp` take network distance/travel time as impedance? 

Yes. I will post an example if I have time tomorrow. I can't share the examples I have to hand 😄 

[...]

PR #465, opened 2024-11-18
add initial commit for routing engine

ljwolf:
This adds the initial routing engine explained in #464.

### DIFF
--- a/spopt/route/engine.py
+++ b/spopt/route/engine.py
@@ -1,0 +1,224 @@
+try:
+    import osrm
+    has_bindings = True
+except (ImportError,ModuleNotFoundError) as e:
+    has_bindings = False
+import os
+import numpy
+import requests
+import warnings
+import geopandas
+import shapely
+from sklearn import metrics
+
+# TODO: needs to be configurable by site
+_OSRM_DATABASE_FILE = ""
+
+def build_route_table(demand_sites, candidate_depots, cost='distance', http=not has_bindings, database_path=_OSRM_DATABASE_FILE, port=5000):
+    """
+    Build a route table using OSRM, either over http or over py-osrm bindings
+    """
+    if isinstance(demand_sites, (geopandas.GeoSeries, geopandas.GeoDataFrame)):
+        demand_sites = demand_sites.geometry.get_coordinates().values
+    if isinstance(candidate_depots, (geopandas.GeoSeries, geopandas.GeoDataFrame)):
+        candidate_depots = candidate_depots.geometry.get_coordinates().values
+    if cost not in ("distance", "duration", "both"):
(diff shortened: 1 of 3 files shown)


notebooks/route.ipynb
jGaboardi, line 1: A couple thoughts from the notebook before getting into code itself

* notebook refers to `spopt.Route`
  * Is that meant to be referring to `route.heuristic.LastMile`?
  * Should we keep `LastMile` or consider renaming to `Route`?
  * Or other ideas?
* need to remove hardcoded pathing -- `dylan/projects/gsoc2025`

jGaboardi: @fiendskrah 

> * need to remove hardcoded pathing -- dylan/projects/gsoc2025

Let me know once you've updated the pathing

> * notebook refers to spopt.Route
>   * Is that meant to be referring to route.heuristic.LastMile?
>   * Should we keep LastMile or consider renaming to Route?
>   * Or other ideas?

What are your thought here?

fiendskrah: > A couple thoughts from the notebook before getting into code itself
> 
> * notebook refers to `spopt.Route`
>   
>   * Is that meant to be referring to `route.heuristic.LastMile`?
>   * Should we keep `LastMile` or consider renaming to `Route`?
>   * Or other ideas?
> * need to remove hardcoded pathing -- `dylan/projects/gsoc2025`

Sorry I did not see this back when you originally posted it. 

> * need to remove hardcoded pathing -- `dylan/projects/gsoc2025`

I have removed the hardcoded pathing issue and will push it up momentarily. 

>   * Should we keep `LastMile` or consider renaming to `Route`?

I would posit that `Route` is a module of `spopt`, where `LastMile` is the first of potentially many solver classes within Route, though I'd defer to the lead maintainers about this.

jGaboardi: > I would posit that `Route` is a module of `spopt`, where `LastMile` is the first of potentially many solver classes within `Route`, though I'd defer to the lead maintainers about this.

So that would be the `route.py` module, but that seems OK from my view point.

jGaboardi: > I have removed the hardcoded pathing issue and will push it up momentarily.

Ping me once you have pushed up those removals.

fiendskrah: I pushed the notebook change up to my [PR with the test suite](https://github.com/ljwolf/spopt/pull/7), merging that would propagate it here as well @ljwolf. Let me know if you'd prefer it into pysal:main separately

jGaboardi: I think we'd like that to propagate here as well. But we'll wait for @ljwolf's direction.

PR #482, opened 2025-04-18
ENH: add Spatial Adaptive Agglomerative Aggregation (SA3) regionalisation algorithm

u3ks:
Hi all,

We implemented an algorithm to delineate contiguous areas within cities that have identical characteristics and configurations of buildings and streets, but we  thought it might be useful for other applications since the procedure is quite generic. 

The idea is that is a kind of spatially restricted HDBSCAN, so there is only one parameter to specify - the minimum number of observations to form a cluster. The procedure basically consists of two steps: first, carrying out a full spatially, restricted ``sklearn.cluster.AgglometariveClustering`` clustering; and second, extracting clusters from the resulting linkage matrix, using density-clustering extraction algorithms - Excess of Mass or Leaf. This results in multiscale (clusters have varying ranges of internal similarity), contiguous clusters with noise (some observations are not attached to any clusters).

I try to explain more how it works, examples and advantages and disadvantages in the ``sa3.ipynb`` notebook.


spopt/region/sa3.py
```diff
@@ -0,0 +1,221 @@
...
+        Cluster labels for observations.
+
+    """
+
+    def __init__(
+        self,
+        gdf,
+        w,
+        attrs_name,
+        min_cluster_size=15,
```
martinfleis, line 106: The default is meaningless without knowing the use case. Shall we maybe make it a required arg?

martinfleis:
@jGaboardi can we bump lipysal min req to 4.10 here?

jGaboardi:
> @jGaboardi can we bump lipysal min req to 4.10 here?

Let's open an issue for that and discuss with @knaaptime, @gegen07, @ljwolf. I think it will be OK, but want to get their inputs.

martinfleis (approved):
This is now fine by me! @u3ks I pushed some changes to future-proof the API (`extraction` keyword that is not a bool) and to clean the API (kwargs passed directly to sklearn rather than via a dedicated dictionary).

knaaptime:
cool

knaaptime (approved):

gegen07 (approved):
LGTM

PR #486, opened 2025-05-08
bump libpysal requirement to 4.10

### LINKED ISSUE #485: Bump libpysal req to 4.10
In #482 we are relying on `Graph` and specifically on `Graph.subgraph`, which needs libpysal 4.10. Any objection to bumping minimal requirement? I am aware it does not necessarily follow SPEC0 but my understanding is that within PySAL we don't need to follow it.

martinfleis:
Closes #485

### DIFF
--- a/ci/311-oldest.yaml
+++ b/ci/311-oldest.yaml
@@ -9,7 +9,7 @@ dependencies:
   - networkx=3.2
   - numpy=1.23
   - pandas=2.0
-  - scikit-learn=1.3
+  - scikit-learn
   - scipy=1.11
   - shapely=2.0.1
   - spaghetti=1.6.4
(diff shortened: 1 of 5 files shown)


martinfleis:
Yeah but it seems that this generally needs newer deps than SPEC0 so not sure if it's worth doing this bump.

jGaboardi:
> Yeah but it seems that this generally needs newer deps than SPEC0 so not sure if it's worth doing this bump.

Is `fast_hdbscan` not now a requirement? Maybe I overlooked that.

jGaboardi:
> Yeah but it seems that this generally needs newer deps than SPEC0 so not sure if it's worth doing this bump.

Let me know when you make a decision.

jGaboardi:
> To remove all the version flags from $SA3$ , we would have to use sklearn 1.6, which is the latest. I can still bump libpysal and add a version check to CI to skip if older, if you think it makes sense. But it won't work with the oldest anyway.

I think this is a good compromise. Go for it.

jGaboardi (approved):

PR #487, opened 2025-06-03
Add Flow Refueling Location Model (FRLM)

fengzixin0617:
- Add FRLM class and example notebook
- With exact and greedy solver options 
- Combined capacited FRLM and basic FRLM


spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+        self.e_coefficients[od_pair] = e_q
+        return e_q
+    
+    def compute_facility_usage(self, 
+                              origin: Any, 
+                              destination: Any, 
+                              facility: Any, 
+                              combination: List) -> int:
+        """
+        Compute facility usage coefficient for capacitated model. (0, 1, or 2)
```
ljwolf, line 313: What does this mean?

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+            self.compute_refueling_frequency(od_pair[0], od_pair[1])
+        
+        self.path_refueling_combinations = path_refueling_combinations
+        
+        total_valid = sum(len(combinations) for combinations in path_refueling_combinations.values())
+        print(f"Generated {total_valid} valid combinations across {len(path_refueling_combinations)} OD pairs")
+        
+        return path_refueling_combinations
+    
+    def calculate_node_weights(self, method: str ='double') -> np.ndarray:
```
ljwolf, line 393: Is `double/single` what's used in the literature? I think this should be implemented as `include_destination=False`.

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+         max_iterations: int = 100,
+         **kwargs) -> Dict:
+        """
+        Solve the FRLM problem with specified solver.
+        
+        Parameters:
+        -----------
+        solver : str, default="greedy"
+            Solver to use: "greedy", "pulp", "exact"
+        solver_instance : Optional[Any], default=None
```
ljwolf, line 442: No. This should be just done using `solver`. if `solver == "greedy"`, we dispatch to `_solve_greedy()`. otherwise, we assume this is a `solver_instance`, and we send it to `self._solve_pulp()`. 

The type is `Union[str, pulp.Solver]`.

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+    def _solve_greedy(self, 
+                 objective: str = "flow",
+                 weight_method: str = "double",
+                 seed: Optional[int] = None,
+                 initialization_method: str = "empty",
+                 max_iterations: int = 100,
+                 **kwargs) -> Dict:
+        
+        import random
+        import numpy as np
```
ljwolf, line 499: Why are these imported here?

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+
+        self.solver_type = "greedy"
+
+        if seed is not None:
+            random.seed(seed)
+            np.random.seed(seed)
+            if kwargs.get('verbose', True):
+                print(f"Random seed set to: {seed}")
```
ljwolf, line 507: Use `logging.log()` if you want to keep track of this, not print.

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+            if iteration_count > max_iterations:
+                if kwargs.get('verbose', True):
+                    print(f"Reached maximum iterations ({max_iterations}), stopping early")
+                break
+            
+            best_facility = None
+            best_value = 0
+            iteration_candidates = {}
+            
+            for i in range(1, num_facilities + 1):
```
ljwolf, line 661: The logic of this solution technique needs to be explained.

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+                   weight_method: str = "double",
+                   **kwargs) -> Dict:
+
+        start_time = time.time()
+        self.solver_type = "pulp"
+        
+        if self.capacitated:
+            return self._solve_capacitated_pulp(solver_instance, max_time_seconds, objective, **kwargs)
+        else:
+            return self._solve_basic_pulp(solver_instance, max_time_seconds, weight_method, **kwargs)
```
ljwolf, line 1036: Again, these should be one method with one `if` statement depending on `self.capacity is None`.

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+            
+            plt.axis('off')
+            
+            if output_file:
+                plt.savefig(output_file, bbox_inches='tight')
+                print(f"Visualisation saved to {output_file}")
+            else:
+                plt.show()
+                
+        except ImportError:
```
ljwolf, line 1550: Try to import, then raise `ImportError('matplotlib is required to use this function')`. Don't put all the code within the `try:` block---just the import!

```python
try:
    import matplotlib.pyplot as plt
except (ImportError, ModuleNotFoundError) as e:
    raise type(e)('Matplotlib is required to use this plotting function.')
```

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+                            uncovered = self.flows[od_pair] * (1.0 - coverage.get('covered_proportion', 0.0))
+                            improvement += uncovered
+                
+                best_improvement = max(best_improvement, improvement)
+        
+        return best_improvement
+
+    @property
+    def pulp(self):
+        return self.problem
```
ljwolf, line 1933: Is this necessary? why not use `self.problem`?

spopt/locate/frlm.py
```diff
@@ -0,0 +1,2089 @@
...
+        print(f"Total variables: {len(self.variables)}")
+
+    
+    @property
+    def problem(self):
+        if self.solver_type == "pulp":
+            return self.model 
+        elif self.solver_type == "greedy":
+            return self._create_greedy_problem_view()
+        return None
```
ljwolf, line 1957: This `return None` should raise an error instead, since it's an mis-configured problem.

spopt/locate/frlm.py
fengzixin0617: I’ve temporarily removed the plot function while I try to figure out a more flexible approach to incorporate different types of networks (e.g., libpysal graphs, networkx, sparse graphs).

spopt/locate/flow.py
```diff
@@ -0,0 +1,2579 @@
...
+        return self.variables.get(var_name)
+
+    def get_constraint_dual(self, constraint_name: str) -> Optional[float]:
+        return self.lagrange_multipliers.get(constraint_name)
+
+    def get_reduced_cost(self, var_name: str) -> Optional[float]:
+        return self.reduced_costs.get(var_name)
+    
+    @classmethod
+    def convert_to_scipy_sparse(cls, network):
```
gegen07, line 143: Is this a classmethod? A classmethod should return a `FRLM` instance, not different instances of a graph. I think it is better to include this in `util.py`

spopt/locate/flow.py
```diff
@@ -0,0 +1,2579 @@
...
+            model += (
+                pulp.lpSum(
+                    y[(q, h)]
+                    for h in range(len(self.path_refueling_combinations[od_pair]))
+                    if (q, h) in y
+                )
+                <= 1
+            )
+
+        model.solve(pulp.GUROBI_CMD(msg=False))
```
gegen07, line 1858: change for `pulp.solverinstance`

spopt/locate/flow.py
```diff
@@ -0,0 +1,2579 @@
...
+                            node_importance[node] += volume
+
+            sorted_nodes = sorted(
+                self.candidate_sites, key=lambda x: node_importance[x], reverse=True
+            )
+            init_count = min(2, len(sorted_nodes))
+
+            return set(sorted_nodes[:init_count])
+
+    def get_solver_details(self, verbose: bool = True) -> Dict:
```
gegen07, line 1951: yes, the detailed information from solver should look like this method

spopt/locate/flow.py
```diff
@@ -0,0 +1,2579 @@
...
+                        "objective_type": objective,
+                    }
+                )
+
+            return result
+        else:
+            print(f"No optimal solution found. Status: {self.status}")
+            return {"status": self.status}
+
+    def _evaluate_capacitated_solution(self, facilities: Dict[Any, int]) -> Dict:
```
gegen07, line 1811: @fengzixin0617 I did not understand this method. We are already defining the capacity constraints within `solve_pulp` method but here are we defining it again?

gegen07 (commented):
Great work! Some comments about the code. I have not reviewed the whole file but I think it is good for this round. :)

spopt/locate/flow.py
```diff
@@ -0,0 +1,2579 @@
...
+        ----------
+        network : object
+            Network representation to be converted to scipy sparse matrix
+        
+        Returns
+        -------
+        self : FRLM
+            Returns the instance for method chaining
+        """
+        self.scipy_network = self.convert_to_scipy_sparse(network)
```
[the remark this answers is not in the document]
ljwolf: Yep! We should just assume input networks area already `scipy.sparse` arrays

spopt/locate/util.py
```diff
@@ -1,7 +1,14 @@
 import geopandas
 import numpy
 from shapely import MultiPolygon, Point, Polygon
-
+from libpysal.graph import Graph
+import networkx as nx
+import libpysal
+import numpy as np
+import scipy.sparse as sp
+from typing import Union
```
gegen07, line 9: Check imports not used

spopt/locate/flow.py
```diff
@@ -0,0 +1,2630 @@
...
+            setattr(obj, "node_coverage_vars", node_coverage_vars)
+        else:
+            raise AttributeError(
+                "Facility and flow variables must be set before adding threshold constraints."
+            )
+   
+    @staticmethod
+    def add_combination_variables(
+        obj: Any,
+        var_name: str = "v_{h}"
```
gegen07, line 378: variable not used and should be a f-string `...=f"v_{h}"`. Check others variable names

spopt/locate/flow.py
```diff
@@ -0,0 +1,2630 @@
...
+        for od_pair, path in self.flow_paths.items():
+            distance = 0
+            for i in range(len(path) - 1):
+                try:
+                    distance += scipy_network[path[i], path[i + 1]]
+                except:
+                    distance += 1.0
+            self._path_distances[od_pair] = distance
+
+    def _calculate_covered_nodes(self) -> List[int]:
```
gegen07, line 2065: private method not used

spopt/locate/flow.py
```diff
@@ -0,0 +1,2630 @@
...
+        for od_pair, coverage in self.flow_coverage.items():
+            coverage_data.append({
+                "origin": od_pair[0],
+                "destination": od_pair[1],
+                "flow_volume": coverage["flow_volume"],
+                "covered_proportion": coverage["covered_proportion"],
+                "covered_volume": coverage["covered_volume"],
+            })
+        coverage_df = pd.DataFrame(coverage_data)
+        coverage_df.to_csv(filename.replace(".csv", "_coverage.csv"), index=False)
```
gegen07, line 2175: We can let the user decide which format of document they would like, following pandas DataFrame specification. Then we can just return a dataframe, and the user can convert to a document.

spopt/locate/flow.py
```diff
@@ -0,0 +1,2630 @@
...
+                            uncovered = self.flows[od_pair] * (
+                                1.0 - coverage.get("covered_proportion", 0.0)
+                            )
+                            improvement += uncovered
+
+                best_improvement = max(best_improvement, improvement)
+
+        return best_improvement
+
+    def get_shadow_prices(self, verbose: bool = True) -> Dict:
```
gegen07, line 2445: `verbose` variable is not used

spopt/locate/flow.py
```diff
@@ -0,0 +1,2630 @@
...
+                "pulp_status": self.pulp_status,
+                "model_name": self.model.name if self.model else 'N/A',
+                "solver_stats": self.solver_stats,
+                "lagrange_multipliers": dict(list(self.lagrange_multipliers.items())[:5]),
+                "reduced_costs": dict(list(self.reduced_costs.items())[:5])
+            })
+
+        return solver_details
+
+    @property
```
gegen07, line 2134: why use this as a property?

spopt/locate/flow.py
```diff
@@ -0,0 +1,2630 @@
...
+        self.dj = 0.0
+
+
+class GreedyConstraint:
+    def __init__(self, name, info):
+        self.name = name
+        self.pi = info.get("shadow_price", 0.0)
+        self.slack = info.get("slack", 0.0)
+
+class FlowModelBuilder:
```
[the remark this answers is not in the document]
qszhao: Great observation, Germano - I wonder how many general variables such as add_facility_variables exist in FacilityModelBuilder and FlowModelBuilder? If we only have limited overlap, then I think we can still keep them separated. Flow models are certainly different from our previous facility models. We are covering arcs by using nodes, rather than using nodes to cover nodes or allocate nodes to each other in the median problem.

gegen07 (approved):
LGTM 🎩

Issue #494, opened 2025-06-15
reconsider how we are storing testing data


jGaboardi:
* reconsider how we are storing [testing data](https://github.com/pysal/spopt/tree/0a5f69caa6a77cca9e3a21f05fb7d4c6b3adebc2/spopt/tests/data)
* currently a mishmash of `.pkl`, `.csv.`, `.shp`
* xref https://github.com/pysal/spopt/issues/370#issuecomment-2974477787

Issue #370, opened 2023-05-07
swap out `simulated_geo_points()` with new `sample_points()` geopandas method


jGaboardi:
We should look at swapping out [`simulated_geo_points()`](https://github.com/pysal/spopt/blob/main/spopt/locate/util.py#L8) with new [`sample_points()` geopandas method](https://geopandas.org/en/stable/docs/changelog.html#version-0-13-may-6-2023).

jGaboardi:
Implementing this will require a `geopandas==0.13.0` stipulation, so we'll have to hold off for a bit.

PR #495, opened 2025-07-06
store tests & testing data by subpackage 

jGaboardi:
* hierarchical testing schema
* simply moving files into subdirs by subpackage - no substantive change
* prep for #494 
  * possibly some test data generator functionality for ease of recreation
* xref #370


gegen07 (approved):
Nice catch! Then the test suite for route would be located in `spopt/tests/test_route`.

PR #500, opened 2025-10-05
Fix PULP default solver installation

gegen07:
Related to [#806](https://github.com/pysal/libpysal/pull/806) and #348.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,6 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - python=3.13
+  - coin-or-cbc
   - fast_hdbscan
   - geopandas
   - jupyterlab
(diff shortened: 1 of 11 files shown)


gegen07:
I think it is hanging on [`test_combination_approach`](https://github.com/pysal/spopt/blob/main/spopt/tests/test_locate/test_flow.py#L72) of FRLM. This method is dependent on `rising_combination`(https://github.com/pysal/spopt/blob/main/spopt/locate/util.py#L109) which generates combinations for refueling. This sounds weird to me, why `COIN_CMD` in UNIX-based can solve this problem and windows cannot.

jGaboardi:
> This sounds weird to me, why COIN_CMD in UNIX-based can solve this problem and windows cannot.

Exactly... strange indeed...

jGaboardi:
@gegen07 – the initial problem causing this seems to have been a fluke and/or it was worked out upstream. Now things seems to be passing, right?

Shall we close this PR?

gegen07:
Seems ok to me. Closing it.

Issue #481, opened 2025-04-18
Suggestion: Add `CONTRIBUTING.md` for smoother onboarding


Abhishek-Dimri:
Hi @jGaboardi,

As a new contributor, I found it a bit tricky to get everything up and running quickly - especially with optional dependencies and development setup. I’d like to suggest adding a `CONTRIBUTING.md` file to help others get started more smoothly.

I did find the [Installation page](https://pysal.org/spopt/installation.html) - which is helpful - but it would be great to have a more centralized place with quick info for contributors, like how to install everything in one go or how to run tests.

If any of the following already exist, please let me know - I might've just missed them!

### Suggested Outline for `CONTRIBUTING.md`

- **Developer’s Guide** - Quick intro for contributors.
- **Requirements** - Python 3.11+, pip or conda.
- **Getting Started** - Fork, clone, and add upstream.
- **Installation** - Something like `pip install .[docs,tests,dev,notebooks]` to install all dev dependencies easily.
- **Running Tests** - Mention how to run tests with `pytest`.
- **Code Style** - Not sure if tools like `ruff`, `black`, or `pre-commit` are used - if yes, would be helpful to document.
- **Documentation** - How to build with Sphinx (`make html`).
- **Notebooks** - There are some optional notebook dependencies, so info on how to run those would help.
- **Making a Release** - For maintainers (build, version bump, publish).
- **Contributing Tips** - If needed.

Happy to help draft a basic version if this sounds useful!

Thanks 🙌

jGaboardi:
@Abhishek-Dimri Thanks for opening the issue!

While we do have a [`CONTRIBUTING.md `](https://github.com/pysal/spopt/blob/main/.github/CONTRIBUTING.md) with some of this information in there and pointer links to other resources (pointed to from [*Contribute* in our `README.md`](https://github.com/pysal/spopt?tab=readme-ov-file#contribute)), it could certainly use some updating.

Moreover, [*Documentation/Developer*](https://pysal.org/docs/devs/) at `pysal.org` could use some love.

cc @gegen07 @knaaptime @sjsrey @martinfleis

PR #505, opened 2026-01-08
Improve CONTRIBUTING.md for new contributors

AKSrivastava28:
This PR improves CONTRIBUTING.md by adding clearer guidance for:
- Local development setup
- Running tests
- Building documentation

This follows up on the discussion in issue #481.


martinfleis:
@gegen07 I am bit hesitant in merging individual changes across the federation as it means each sub package will have its own guidelines, complicating federation-wise maintenance. I think we should think about how to standardise this.

PR #511, opened 2026-03-24
ENH: warn and fall back to p-median when k    >= p_facilities in KNearestPMedian

### LINKED ISSUE #428: KN-PMP: solve standard PMP once any `k>=p`
Currently within the $k$-nearest $p$-median problem,  we raise errors here[^1] and here[^2] when any values of `k` is greater than `p`. Perhaps we should keep this behavior in `from_geodataframe()`, but in `_create_sparse_matrix()` we should probably consider throwing a warning then defaulting to solve a standard $p$-median problem.

[^1]: in [`_create_sparse_matrix()`](https://github.com/pysal/spopt/blob/bf7134a33794e9a02377e95563faaa48d74a486d/spopt/locate/p_median.py#L723)
[^2]: in [`from_geodataframe()`](https://github.com/pysal/spopt/blob/bf7134a33794e9a02377e95563faaa48d74a486d/spopt/locate/p_median.py#L967)

xref #425

CONFUZ3:
Closes #428

  ## What changed
  - `KNearestPMedian.from_geodataframe()` now emits a `UserWarning` when any value in `k_array >= p_facilities`, explaining that the k-nearest constraint is non-binding and the model degenerates to astandard p-median.
  - `k_array` is reset to `n_facilities` for all clients to make the equivalence explicit.
  - Added two new tests covering `k == p` and `k > p` cases.
  - Updated `test_error_high_capacity` to also assert the new warning fires (that test used `p_facilities=1` with `k=[1,1]`).

  ## Why `from_geodataframe` not `_create_sparse_matrix`
  `solve()` calls `_create_sparse_matrix()` in a loop, and `_update_k_array()` naturally increments k values during iteration; checking there would produce spurious warnings. Guarding at construction time catches only user-supplied initial k values.

### DIFF
--- a/spopt/locate/p_median.py
+++ b/spopt/locate/p_median.py
@@ -977,6 +977,21 @@ class KNearestPMedian(PMedian):
                 f"of total facilities, which is {len(fac_data)}."
             )
 
+        # When any k >= p_facilities the k-nearest constraint is non-binding:
+        # only p facilities are ever selected, so a client with k >= p can always
+        # reach every selected facility. The model degenerates to a standard
+        # p-median; warn and set k = n_facilities to make the equivalence explicit.
+        if (k_array >= p_facilities).any():
+            warnings.warn(
+                "Some ``k`` values are >= ``p_facilities`` "
+                f"({p_facilities}); the k-nearest constraint is "
+                "non-binding and the model degenerates to a standard "
+                "p-median. Solving as a standard p-median instead.",
+                UserWarning,
+                stacklevel=2,
+            )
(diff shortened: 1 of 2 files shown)


spopt/locate/p_median.py
```diff
@@ -977,6 +977,21 @@ def from_geodataframe(
                 f"of total facilities, which is {len(fac_data)}."
             )
 
+        # When any k >= p_facilities the k-nearest constraint is non-binding:
```
jGaboardi: Condense this 4-line comment down to 1 line please.

PR #512, opened 2026-03-25
Use non-strict threshold when identifying candidate move areas

hegen008:
The max-p-regions problem uses a non-strict threshold, requiring that regions have a spatially extensive attribute greater than or equal to a predetermined threshold. This PR replaces > with >= in the pick_move_area() function to align with the non-strict threshold definition. With this improvement, when areas are identified as candidates to be moved in the simulated annealing phase, the remaining parts of the region are required to be >= the minimum region threshold, rather than strictly greater than it. The simulated annealing algorithm works with a strict threshold as originally written, but the results will generally not be as good with a lower number of potential move candidates identified. As this improvement affects the simulated annealing algorithm path, the expected result for 2 test cases needed to be updated as well.

### DIFF
--- a/spopt/region/maxp.py
+++ b/spopt/region/maxp.py
@@ -498,7 +498,7 @@ def _pick_move_area(
         rla = np.array(region_lists[k])
         rasa = threshold_array[rla]
         lost_sa = v - rasa
-        pas_indices = np.where(lost_sa > threshold)[0]
+        pas_indices = np.where(lost_sa >= threshold)[0]
         if pas_indices.size > 0:
             for pasi in pas_indices:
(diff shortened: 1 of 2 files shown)


knaaptime:
thanks!

yeah, i think that's right, following the original JRS paper
Screenshot 2026-03-25 at 11 30 41 AM

i'll let @sjsrey weigh in to be absolutely sure though

hegen008:
Thanks for confirming that the problem definition has a non-strict threshold. To clarify to purpose of this PR, the algorithm is currently implemented with a non-strict threshold in most places, it is just this one line that uses a strict threshold instead. I think what is most important is that whichever is chosen is used consistently.

PR #516, opened 2026-04-11
fix: validate SA3 extraction mode early

samay2504:
This PR fixes the SA3 extraction error path by validating the extraction mode before importing optional clustering dependencies, so unsupported values now raise the expected `ValueError` immediately instead of being masked by an `ImportError` when `fast_hdbscan` is unavailable; this keeps user-facing errors accurate and aligns runtime behaviour with the existing test expectation.

### DIFF
--- a/spopt/region/sa3.py
+++ b/spopt/region/sa3.py
@@ -32,6 +32,9 @@ def extract_clusters(linkage_matrix, min_cluster_size, extraction="eom"):
 
     """
 
+    if extraction not in ("eom", "leaf"):
+        raise ValueError("Unsupported extraction method. Use one of ('eom', 'leaf').")
+
     n_samples = linkage_matrix.shape[0] + 1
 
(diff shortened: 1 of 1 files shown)


martinfleis:
How does it matter which error bubbles up first? You need both - correct method and dependency.

samay2504:
Good point, thanks. The reason I changed the order is that extraction validation is independent of optional deps, so a typo in extraction currently surfaces as a dependency error and can send users in the wrong direction. My intent was to fail fast on invalid user input first, then only check optional deps for valid extraction paths. That said, I can absolutely switch this if you prefer dependency-first error bubbling here. Happy to update to your preferred behaviour.

martinfleis:
It simply does not matter. If you make a typo, fixing it does not help you if you don't have the dependencies. And there's no path in SA3 that would avoid using them.

Issue #524, opened 2026-04-11
FRLM get_flow_coverage return annotation is incorrect


samay2504:
## Description
`FRLM.get_flow_coverage()` returns a coverage dictionary but is currently annotated as returning `None`, which is inconsistent with runtime behaviour and makes downstream type checking and editor hints inaccurate.

## Steps to reproduce
```python
coverage = model.get_flow_coverage()
print(type(coverage))
```

## Expected behaviour
The method signature should declare a dictionary return type that matches the returned object.

## Actual behaviour
The method is annotated with `-> None` even though it returns a dictionary.

## Environment
- spopt version: main
- Python version: 3.11
- OS: Windows 11

PR #525, opened 2026-04-11
fix: correct FRLM flow coverage annotation

samay2504:
This PR fixes issue #524 by updating `FRLM.get_flow_coverage()` to use a dictionary return annotation that matches its actual return value, and it adds a small test assertion that checks the returned object type, which keeps the public API contract explicit for type checkers and editor tooling without changing runtime behaviour.

### DIFF
--- a/spopt/locate/flow.py
+++ b/spopt/locate/flow.py
@@ -373,7 +373,7 @@ class FlowModelBuilder:
 class FRLMCoverageMixin:
     """Mixin to calculate flow coverage statistics for FRLM."""
 
-    def get_flow_coverage(self) -> None:
+    def get_flow_coverage(self) -> dict[str, float]:
         """Improved and fixed calculation of flow coverage."""
         if not hasattr(self, "facility_vars") or self.facility_vars is None:
             raise AttributeError(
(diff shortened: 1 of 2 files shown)


jGaboardi:
@samay2504 – You have opened more than 5 Issue+PR pairs in a very short period. Please provide a statement on each associated PR as to the degree of LLM usage that was utilized to both create the issue & PR. This statement should be written by you, not CoPilot or any other LLM.

I can't speak for other maintainers, but I am unwilling to review mass-produced PRs generated by LLMs.

If no explanation is provided in a timely manner, we reserved to right to close them.

cc @knaaptime @martinfleis

samay2504:
For transparency, I prepared and validated this change myself; I identified the mismatch between the annotation and runtime return type, made the code and test edits locally, ran the checks, and reviewed the diff before opening the PR. I did use tooling assistance in a limited way for wording and minor drafting support, but the technical diagnosis, implementation decisions, testing, and final commit content were done by me. I understand the concern about high-volume submissions, and I will slow down my pace and include explicit disclosure notes on each related PR going forward.

